PR #19, opened 2018-06-04
Add missing NIMBUS scalarising functions

frankier:
Also, split achivement function into two one for NIMBUS and one for NAUTILUS.


frankier:
I've checked/fixed the Nautilus scalarisation function by running article_nautilus. I don't know what the correct output is, but I see no obvious problems so I've merged this now.

PR #38, opened 2018-11-23
Comments

podkop:
First test pull request in order to learn how it works.
Added some comments to "Problem"

### DIFF
--- a/desdeo/problem/Problem.py
+++ b/desdeo/problem/Problem.py
@@ -15,7 +15,10 @@ from desdeo.utils.misc import as_minimized
 
 class MOProblem(ABC):
     """
-    Abstract base class for multiobjective problem
+    Abstract base class for a multiobjective minimization problem,
+    which represents some instance of an original multiobjective problem.
+        If the original problem contains maximization objectives, 
+        they should be multiplied by -1 before implementation.
 
     Attributes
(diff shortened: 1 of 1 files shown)


ferrety:
Hmm, there seems to be new [build errors](https://travis-ci.com/industrial-optimization-group/DESDEO/pull_requests) in addition to our existing ones. 

DESDEO enforces syntax and [pep8](https://www.python.org/dev/peps/pep-0008/) checks. Developers should install pre commit  [hooks as per docs](https://github.com/industrial-optimization-group/DESDEO#development) to check them before pushing to github. You should also run `pytest tests` in order ensure unittests pass (such as they are) before committing.

podkop:
If you noticed I only change comments to the code. The "comments" branch "master" forked for discussion, extending docs and asking you questions.

ferrety:
> 
> If you noticed I only change comments to the code. The "comments" branch "master" forked for discussion, extending docs and asking you questions.

Editing the comments can also break the builds, as they might violate pep8. You can see from the error logs that changes add trailing white spaces. You can see the lines of code where the errors are from the log. 

The failing unittest is something that needs to be fixed, as it should mean that the current version does not work (or that the unittest is incorrect). Might be due updated depency.

PR #40, opened 2019-03-18
decision uncertainty example

yuezhoukangas:
Added an example and a toy problem (based on river pollution) for solving problems with decision uncertainty.


yuezhoukangas:
All checks failed as shown above. Running pytest tests did not give problem. Trying to configure pre-commit to run python3.7 did not help.

EchoBetaGlass:
Builds have been failing for quite some while now. I will look into this problem when I switch my full focus to this repository.

Issue #69, opened 2023-12-02
[IDEA] Problem Formulation


light-weaver:
* **What is the current behavior?**
Problems are formulated using the MOProblem class.

* **Describe the solution you'd like**
Problem formulations should be represented as JSON objects. They can be read into Python as Dataclasses and stored in databases without changes. This requires analytical formulations of objectives to be stored as MathJSON objects instead of Numpy expressions. The MathJSON objects can be converted to Polars expression for evaluation with currently implemented methods. Alternatively, we can implement other converters that convert the problem formulation to, for example, numpy/pandas expressions, PuLP expressions, Gurobipy expression. We can even convert the MathJSON objects to industry-standard file formats (only for single objective optimization. [Yan's work](https://github.com/industrial-optimization-group/desdeo-problem/pull/90) is the first step towards this idea. 

* **What is the motivation/use case for changing the behavior?**
Currently, arbitrary python objects have to be stored into the database. This is bad behaviour and prevents complicated use cases such as updating/changing problem formulation.

* **Additional context**
Insight into how to handle surrogate modelling, external simulators, arbitrary binaries, and scenario based optimization needs further discussion.

gialmisi:
I have updated the pydantic schema for representing problems in DESDEO. It can be viewed [here](https://desdeo.readthedocs.io/en/desdeo2/explanation/problem_format/#the-pydantic-schema) A relational map of the current schema is visualized below:

![image](https://github.com/industrial-optimization-group/DESDEO/assets/50820090/9a09d42b-3058-4f6c-8943-e27d06f4d746)

I have also been thinking of the logic of parsing the JSON representation into an expression that can be evaluated. The flow of the logic has been described [here](https://desdeo.readthedocs.io/en/desdeo2/explanation/problem_format/#parsing). The image illustrating the flow as also shown in this post below.

![image](https://github.com/industrial-optimization-group/DESDEO/assets/50820090/cac2a986-b60c-4d24-ac67-3ea7748cb0a6)

At least the schema should cover analytical problems, but the question of surrogate-based problems remains open. In any case, we can add to the schema going forward without having to break any previous functionalities.

gialmisi:
@light-weaver I updated the parser logic. Do you think `ScalarizationFunction` in the problem schema should also have a symbol? Initially, I assumed this would not be needed (we usually do not re-use scalarization function values in other functions), but now that I think of it, this could be useful in, e.g., PIS-based methods?

[...]

PR #77, opened 2023-12-14
Light-weaver/issue69

light-weaver:
Initial schema example


light-weaver:
Initial schema example to fix #69

PR #86, opened 2024-03-26
Update scipy_solver_interfaces.py

juropo:
I was looking through the code to figure out how I would go about adding support for a gurobi solver, and this line seemed wrong.  Please check if this line should reference problem.constraints instead of problem.constants

### DIFF
--- a/desdeo/tools/scipy_solver_interfaces.py
+++ b/desdeo/tools/scipy_solver_interfaces.py
@@ -265,7 +265,7 @@ def create_scipy_minimize_solver(
 
     def solver(target: str) -> SolverResults:
         # add constraints if there are any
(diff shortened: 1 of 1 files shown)


gialmisi:
Thanks @juropo 

Indeed, this is a mistake. 

If you are looking into how to add Gurobi support, I suggest you look into the Bonmin interface [here](https://github.com/industrial-optimization-group/DESDEO/blob/desdeo2/desdeo/tools/pyomo_solver_interfaces.py) It is probably best to implement Gurobi support through pyomo models. Depending on how your problem is specified, we might also need to look into more ways of setting up a `Problem` model. I am facing this same issue soon as well, since I have a parameter-heavy combinatorial problem I would like to model.

PR #90, opened 2024-04-12
Desdeo2 gurobipy

juropo:
The gurobipy solver and persistent solver should be working now, and there are even some tests.


gialmisi:
Thanks a lot for this! I had no issues merging the changes.

PR #94, opened 2024-04-23
Pareto navigator

Matskuu:
Added implementation for Pareto Navigator and also added pareto_navigator_test_problem as an import from desdeo.problem.


desdeo/mcdm/pareto_navigator.py
```diff
@@ -0,0 +1,271 @@
...
+
+    Returns:
+        dict[str, float]: A reference point converted from classification.
+    """
+    ref = []
+    ideal = objective_dict_to_numpy_array(problem, get_ideal_dict(problem))
+    nadir = objective_dict_to_numpy_array(problem, get_nadir_dict(problem))
+    sol = objective_dict_to_numpy_array(problem, current_solution)
+
+    for i in range(len(pref_info["classification"])):
```
gialmisi, line 42: We should avoid referring to the objectives with an index. We can consider that `pref_info` is in the following format:

```python
pref_info = {"f_1": "<", "f_2": ">", "f_3": "="}
```

This way it is always explicitly clear to which objective function each classification belongs to. This allows to easily refer to the correct ideal value as well. E.g., by using the methods [`Problem.get_ideal_point()`](https://desdeo.readthedocs.io/en/desdeo2/api/desdeo_problem/#desdeo.problem.schema.Problem.get_ideal_point) and [`Problem.get_nadir_point()`](https://desdeo.readthedocs.io/en/desdeo2/api/desdeo_problem/#desdeo.problem.schema.Problem.get_nadir_point).

Can you adjust this function to be compliant with the above suggestion?

desdeo/mcdm/pareto_navigator.py
```diff
@@ -0,0 +1,271 @@
...
+
+    Args:
+        allowed_speeds (np.ndarray): An array of allowed speeds.
+        speed (int): A given speed value where.
+
+    Returns:
+        float: An adjusted speed value calculated from given float.
+            Is between 0 and 1.
+    """
+    return (speed / np.max(allowed_speeds)) / 20
```
gialmisi, line 65: Is the value `20` here necessary to be hard-coded? I suggest it is given as an optional argument here, and its default value can be 20.

desdeo/mcdm/pareto_navigator.py
```diff
@@ -0,0 +1,271 @@
...
+    """Construct the A' matrix in the linear parametric programming problem from the article.
+
+    Args:
+        problem (Problem): The problem being solved.
+        matrix_a (np.ndarray): The A matrix from the polyhedral set equation.
+
+    Returns:
+        np.ndarray: The A' matrix in the linear parametric programming problem from the article.
+    """
+    ideal = objective_dict_to_numpy_array(problem, get_ideal_dict(problem))
```
gialmisi, line 116: See my previous comment. We have methods now for the `Problem` class that return the ideal and nadir points as dicts. Consider refactoring them here.

desdeo/mcdm/pareto_navigator.py
```diff
@@ -0,0 +1,271 @@
...
+
+    solutions: list[dict[str, float]] = []
+    while len(solutions) < num_solutions:
+        solution = calculate_next_solution(problem, d, solution, alpha, matrix_a_new, b)
+        solutions.append(solution)
+    return solutions
+
+# Testing
+if __name__ == "__main__":
+    from desdeo.problem import pareto_navigator_test_problem
```
gialmisi, line 218: Can you move these tests to the folder `tests`? It would be nice if there are some asserts as well, but the bare minimum should be that the code runs at least in some case without errors, like is done here now.

gialmisi (changes requested):

desdeo/mcdm/pareto_navigator.py
Matskuu: Muutin tämän nyt niin, että tuossa pref_info argumentissa tulee dictinä nuo luokittelut.

desdeo/mcdm/pareto_navigator.py
Matskuu: Muutin tuon jakajan arvon valinnaiseksi argumentiksi funktiolle. Ei sinänsä ole tietoa, kuinka paljon sitä täytyy muuttaa eri ongelmille, kun tuosta ei ollut mainintaa itse artikkelissa.

desdeo/mcdm/pareto_navigator.py
Matskuu: Vaihdoin nuo get_ideal_dict() ja nadirin vastaavat Problem.get_ideal_point() muotoon koko tiedostoon.

desdeo/mcdm/pareto_navigator.py
Matskuu: Tein tuosta testauksesta oman tiedostonsa ja lisäsin sinne pari assertia ja ainakin näillä meni testit läpi.

gialmisi (approved):
This looks good now!

PR #103, opened 2024-06-03
Desdeo2 guest login

maihoangbichtram:
Create endpoint for guest login


maihoangbichtram:
Close for now, since it requires more implementations for guests

PR #106, opened 2024-06-05
Desdeo2 join link endpoints

maihoangbichtram:
Endpoints for join link logic

- Create Invite db model to store join links' details
- /invite is to create code to access problems page without any credentials needed
- /login-with-invite is to log the user in with the code provided


light-weaver:
@maihoangbichtram Check the conflict.

PR #111, opened 2024-06-20
Test problem RE21

Matskuu:
Implemented the RE21 test problem in desdeo2 format. Added a page for test problem documentations and added some documentation for the RE21 problem. However, I could not find much information about the problem, so at the moment the description is a bit stale.


desdeo/problem/testproblems.py
```diff
@@ -1045,6 +1045,96 @@ def simple_scenario_test_problem():
...
+        Tanabe, R. & Ishibuchi, H. (2020). An easy-to-use real-world multi-objective
+            optimization problem suite. Applied soft computing, 89, 106078.
+            https://doi.org/10.1016/j.asoc.2020.106078.
+
+        https://github.com/ryojitanabe/reproblems/blob/master/reproblem_python_ver/reproblem.py
+
+    Returns:
+        Problem: an instance of the four bar truss design problem.
+    """
+    # should these be hardcoded here or given as optional arguments for the problem?
```
gialmisi, line 1076: Add these as optional arguments with default values set to the current values given to them.

desdeo/problem/testproblems.py
```diff
@@ -1045,6 +1045,96 @@ def simple_scenario_test_problem():
...
+        upperbound=3*a,
+        initial_value=2*a
+    )
+
+    f_1 = Objective(
+        name="f_1",
+        symbol="f_1",
+        func=f"{l} * ((2 * x_1) + {np.sqrt(2.0)} * x_2 + Sqrt(x_3) + x_4)",
+        objective_type=ObjectiveTypeEnum.analytical,
+        is_convex=True
```
gialmisi, line 1121: Explicitly state for each Objective if it is convex or not, is it linear or not, and is it twice differentiable or not. Double check whether this function is convex or not not (I believe it is concave...).

gialmisi (changes requested):
Please address the two comments I left and update the PR.

juropo:

PR #113, opened 2024-06-25
Test problem RE22

Matskuu:
Added implementation for test problem RE22 with some tests and documentation. Also some minor changes to RE21 documentation.


gialmisi (changes requested):
See comments for #111 and apply where necessary.

juropo:

PR #115, opened 2024-06-25
Test problem RE24

Matskuu:
Added implementation for test problem RE24.


gialmisi (approved):
Good job, these are fine now!

Issue #112, opened 2024-06-22
ASFs can have different "reference point" and more importantly, weights for the max and sum terms.


light-weaver:
https://github.com/industrial-optimization-group/DESDEO/blob/b83c360511de4b53834f4f87d3f399ab0641282b/desdeo/tools/scalarization.py#L209-L210

gialmisi:
Addressed in #117 

@light-weaver closing this for now. If the changes in #117 re not adequate, feel free to re-open this with additional details.

PR #117, opened 2024-06-26
Changes requested in issue #112.

Matskuu:
Changes to generic achievement scalarizing functions requested in issue #112: add_asf_generic_diff and add_asf_generic_nondiff now take a second reference point and second weights to be used in the augmentation term. If a second reference point is not given, no reference point is used in the augmentation term. If second weights are not given, the same weights are used in both the max term and the augmentation term.

### DIFF
--- a/tests/test_scalarization.py
+++ b/tests/test_scalarization.py
@@ -87,7 +87,7 @@ def test_add_asf_nondiff(river_w_fake_ideal_and_nadir):
         assert f"{key}_min" in flatten(problem.scalarization_funcs[0].func)
         assert value in flatten(problem.scalarization_funcs[0].func)
 
-
+@pytest.mark.scalarization
 def test_add_asf_generic_nondiff(river_w_fake_ideal_and_nadir):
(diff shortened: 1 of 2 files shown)


tests/test_scalarization.py
gialmisi, line 1: Please modify the test at least so that you check that you get a different solution when you provide `reference_point` and `weights` only, and when you also provide the weights and reference point for the augmentation term.

Modify also the test function ` test_add_asf_generic_nondiff` with the same idea.

gialmisi (changes requested):
This seems to be fine. I would like the tests to be updated also before accepting this PR.

Matskuu:
I have now updated the test functions for the asf's. The differences for the nondiff version seem to be so small in the test case that I modified the rho value and only compared the decision variable values in the nondiff version testing.

gialmisi (approved):
These are fine now. The point of the tests, especially when solved using scipy, is to check that the scalarizations at least work without throwing any errors. So this is enough for now!

PR #119, opened 2024-07-01
Correct an error in scipy_solver_interfaces.

Matskuu:
Corrected an error stemming from giving initial_guess (x0) to scipy_minimize as a dict instead of a list.


gialmisi (approved):
Indeed, good catch!

PR #121, opened 2024-07-05
Tensor handling for GurobipySolver

Matskuu:
Updated GurobipyEvaluator and GurobipySolver to handle TensorConstants and TensorVariables. TensorVariables are added as gurobipy.MVars and stored in the evaluator as gurobipy model only adds MVars as lists of Var objects. MVars can then be used to do matrix calculation with numpy arrays which means any lists (TensorConstant values) are turned into numpy arrays to be compatible with MVars. Also updated SolverResults to allow optimal variable values to be matrices (lists of lists). Tested the updated GurobipySolver using an implemented test problem. The tests implemented so far are passing fine.


juropo:
Looking good

PR #131, opened 2024-07-23
Scalarization functions for multiple decision makers

Matskuu:
Implemented four scalarization functions for multiple decision makers. So far the tests have mostly been "see if the results make sense when solving the scalarization function". Also compared the results from solving the scalarization functions with one reference point (DM) to the results from the original versions of the scalarization functions.


maihoangbichtram:
`Problem found`: unsupported operand type(s) for +: 'MLinExpr' and 'GenExprMax' (which are found in scalarization funcs)

I added quick fixes to the MAX() function in scalarization funcs for UI testing purpose.

[https://github.com/industrial-optimization-group/DESDEO/compare/desdeo2...maihoangbichtram:DESDEO:group-asf-fix](url)

It's probably hard to see since I accidentally included your changes in that commit. Basically I added 'sub_constraints' for ScalarizationFunction. I don't test it thoroughly if the number is correct but at least UI works now with solving-sub-problems step so I hope those quick fixes can be useful to you.

Matskuu:
> `Problem found`: unsupported operand type(s) for +: 'MLinExpr' and 'GenExprMax' (which are found in scalarization funcs)
> 
> I added quick fixes to the MAX() function in scalarization funcs for UI testing purpose.[Here](https://github.com/industrial-optimization-group/DESDEO/compare/desdeo2...maihoangbichtram:DESDEO:group-asf-fix)
> 
> It's probably hard to see since I accidentally included your changes in that commit. Basically I added 'sub_constraints' for ScalarizationFunction. I don't test it thoroughly if the number is correct but at least UI works now with solving-sub-problems step so I hope those quick fixes can be useful to you.

I don't know if it is necessary to, for example, change the problem schema. I made differentiable versions of the group scalarization functions in the same way they were made for the original ones. I did some testing with them but probably some more testing should be done at some point. The conflicts from these updates come from adding lower and upper bounds for the "alpha" variables for the solvers that require them.

maihoangbichtram:
Scalarizations do not work yet. Now it reports 'unsupported operand type(s) for +: 'MLinExpr' and 'NoneType'', still for self.init_scalarizations. '_alpha' seems to be NoneType. The problem is seen for Forest problem (matrix-based) and NIMBUS test problem (float-based).

I recommend to add some tests for solve_sub_problems too.

Matskuu:
> Scalarizations do not work yet. It reports still for self.init_scalarizations. At least now, I see that not all scalarizations have MAX() removed yet, or 'alpha' declared (eg add_group_nimbus_sf).
> 
> It would be great if you have tests for solve_sub_problems too.

Those new differentiable versions for the scalarization functions, for example add_group_nimbus_sf_diff, are meant to be used when the solver can not handle the max terms (should work fine with the forest problem as well). So, the point is to not change the original function, for example add_group_nimbus_sf, but instead have versions available for those solvers as well in addition to the original ones.

maihoangbichtram:
Ah I see! Since the name of scalarizations is changed, my code makes use of the nondiff ones. All good now with the tests!!

PR #137, opened 2024-08-14
Make DESDEO2 runnable on Rahti

juropo:
I added the necessary s2i scripts to make it possible to build DESDEO2 into an image using s2i.

Also, updated some of the database code.
- The config file is now actually used for database connections. You might need to update it for your local testing.
- Changed how the database is cleared if database already exists when db_init is called. I don't think this will change how it will work for anyone else, but this solution avoids problems with user rights when using pukki database service.


light-weaver:
Feel free to merge it if it works. Mai needs this for the work that she's doing.

PR #140, opened 2024-08-20
Added the data file paths as arguments for the forest problem.

Matskuu:
Added the data file paths as an argument for the forest problem. The data files may be in different directory, so now the paths to those files is given as a list argument for the problem.


juropo:
Giving the files as a list named data is awfully non-descriptive. If you want, you can call the first one simulation_results and the other treatments or treatment_key.

Also, if I accept this pull request the ideal and nadir values will always be wrong if I don't use one of very specific files.

PR #143, opened 2024-08-23
Vector support for Polars evaluator

Matskuu:
Added support for vector valued constants and variables for polars evaluator. No testing has been done for the discrete version as no suitable test problem is available at the moment. At least for now, higher dimensions are not supported (will be made an issue).


gialmisi:
Closing this since it is now addressed in commit 0382979

PR #150, opened 2024-08-29
Gnimbus

maihoangbichtram:
- Add route **gnimbus** for Gnimbus method (iterate, intermediate, solution-vote, ...).
- Add route **gmethod** for general group methods (save-request) : In case in the future, there is another group method implemented.
- New table **GSolutionArchive** which is quite similar to  **SolutionArchive** table. Differences: 
1. column **new** for newly generated solutions (which have to be voted to choose a current solution).
2. column **decision_variables** saves data as json to work with problems with tensor variables.

- In **parse_list_to_mathjson** of problem schema, if the first element is already 'List', parsing is not necessary anymore. In db_init, when saving problems to db, parse_list_to_mathjson has been run so the tensor constants are saved as mathjson in db. When problem is fetched from db, it tries to run parse_list_to_mathjson again.
- In **gurobi**, when there are no solutions, it will throw errors when trying to get vars and constraints from model so if there is no solutions, just simply return empty dict.


maihoangbichtram:
Split into small PRs

PR #151, opened 2024-08-29
Avoid problem duplicates

maihoangbichtram:
In case problem A has role_permission for all DMs, and by mistake dm2 (who is a DM) has access to problem A which is defined UserProblemAccess table
-> Problem A is shown twice 
-> Here I check the id of the fetched problems to avoid returing duplicates to UI

### DIFF
--- a/desdeo/api/routers/problems.py
+++ b/desdeo/api/routers/problems.py
@@ -1,9 +1,7 @@
 from typing import Annotated
-
 from fastapi import APIRouter, Depends, HTTPException
 from pydantic import BaseModel
 from sqlalchemy.orm import Session
-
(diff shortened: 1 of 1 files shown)


jpajasmaa:
What do you mean "and **by mistake** dm2 (who is a DM) has access to problem A" ? Would the need for this PR go away if this **mistake** would not happen?

maihoangbichtram:
For now, there is no checking for duplicate access for roles and user ids (There are 2 ways to grant access, by role and by user id). It's less complicated to add checking for problem id rather than to add checking for duplicate access.

juropo:
This seems good to me.

PR #152, opened 2024-08-29
Add db functions

maihoangbichtram:
- Add update function for session
# Without update func
old_current_solutions = (
      db.query(SolutionArchive)
      .filter(SolutionArchive.problem == problem_id, SolutionArchive.user == user.index, SolutionArchive.current)
      .all()
 )

for old in old_current_solutions:
    old.current = False

# With update func
await db.update(
    update(SolutionArchive)
    .where(SolutionArchive.problem==problem_id, SolutionArchive.current==True)
    .values(current=False)
)

- Make use of **scopefunc** (async_scoped_session). By using current_task as scopefunc, a new session will be created for each new coroutine. -> To avoid coroutine issue
- Separate sessions for different requests to API

### DIFF
--- a/desdeo/api/utils/database.py
+++ b/desdeo/api/utils/database.py
@@ -2,14 +2,15 @@
 
 from os import getenv
 from typing import TypeVar, Dict
+from asyncio import current_task
 
 from sqlalchemy.engine import URL, Result
-from sqlalchemy.ext.asyncio import create_async_engine, AsyncEngine, AsyncSession, AsyncScalarResult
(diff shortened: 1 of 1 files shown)


juropo:
Can you explain what the purpose of these changes is?

maihoangbichtram:
Instead of fetching all records and update each record with desired value, I call only 1 update function to update the values that matches the filter conditions.

Each call to API will now there its own db session, instead of 1 db session for all calls to API.

maihoangbichtram:
PR reconstruct

PR #156, opened 2024-09-02
Solver enums

Matskuu:


juropo:
Took a while for me to take a look at this. It looks mostly good.

Looking at find_compatible_solvers function, I was wondering whether it would be possible to programmatically check if ipopt and/or gurobi are installed and only return the associated pyomo solvers if they are?

PR #161, opened 2024-09-03
Fix parsing error when fetching from db

maihoangbichtram:
Problem: 

- In db_init, when saving problems to db, parse_list_to_mathjson has been run so the tensor constants are already saved as mathjson in db. When problem is fetched from db, it tries to run parse_list_to_mathjson again.

-> In parse_list_to_mathjson of problem schema, if the first element is already 'List', parsing is not necessary anymore.


juropo:
I have been using a similar bit of code in my local branch, which I assume both of us got from Matias. No issues with putting this into the main branch.

PR #162, opened 2024-09-03
Gurobi check number of sols before parsing

maihoangbichtram:
In gurobi, when there are no solutions, it will throw errors when trying to get vars and constraints from model so if there is no solutions, just simply return empty dict.


juropo:
No problems with this.

PR #168, opened 2024-09-04
Fix cors

maihoangbichtram:
UI runs on localhost at port 5173 by default


juropo:
Yeah, no problems with changing this.

PR #180, opened 2024-09-18
Add refresh api

maihoangbichtram:
- Add route `/refresh` to check validity of refresh token and to fetch user's data

- Move content of `func get_current_user()` to `authenticate_token()` for the route `/refresh` to be able to make use of. get_current_user() requires token from fastapi.security.OAuth2PasswordBearer, which is not the case in route `/refresh`


juropo:
This seems fine, but because we don't have tests for the api code, whoever merges this should also check that things work more or less like they should after these changes.

PR #187, opened 2024-11-06
Fixing guess_best_solver

Matskuu:
Added some missing checks for solvers into guess_best_solver and fixed the test for find_compatible_solvers.


juropo:
This seems fine, and I need it, so I'm merging it

PR #192, opened 2024-11-14
Tests for simulator/surrogate evaluator

Matskuu:
Added some tests for the simulator/surrogate evaluator to see that it runs fine for different variations of a problem. The tests have two simulator files (in the data folder because I did not know where to put them) and during the test, two surrogate models are stored in pytest temporary directories and loaded from there.

For the tests, I added a simulator_problem into testproblems that has all analytical, simulator and surrogate based objectives, constraints and extra functions to test that the evaluator works with all of them.

Also made some changes to the evaluator itself; I had to comment out everything that used skops.io (for file formatting for the surrogate models), because the tests gave errors because of it, and used joblib and pickle instead (for now). Also made some fixes so that the evaluator works in different situations.


gialmisi:
Thanks, @Matskuu ! I was running tests locally with your changes, and noticed that the test for the test problem re24 is not passing. Any ideas on why?

Matskuu:
> Thanks, @Matskuu ! I was running tests locally with your changes, and noticed that the test for the test problem re24 is not passing. Any ideas on why?

It seems that the formulation of the second objective of the re24 problem has been changed so that it returns 0 in that example case. I do not remember off the top of my head what the original formulation was in the papers but in the docstring I have written, the original version of the implementation, that the test is based off of, is the one correctly implemented. I don't know why the change has been made so I can't really give more information on this.

light-weaver:
No need to look into the RE24 problem right now. I was the one who changed the implementation. I will discuss this issue on Wednesday. 

Briefly: The implementations of most RE problems are wrong. While these problems originate from constrainted problems in literature, here they should be unconstrained (see [reference implementation](https://github.com/ryojitanabe/reproblems/blob/master/reproblem_python_ver/reproblem.py) and [supplementary](https://github.com/ryojitanabe/reproblems/blob/master/doc/re-supplementary_file.pdf), as well as the original paper).

However, there was a more specific issue with RE24. The second objective is supposed to be the sum of constraint violations:
From the supplementary:
image
From the documentation:
image
Note the missing negation.

There's quite a lot of multiplying by (-1) happening in the implementation because we only support LTE constraints (and this is GTE). However, after checking everything, I think I have now implemented it as intended.

However, I am not able to get the Pareto front they showcased. I do not yet have time to figure out whether the problem implementation is still incorrect, or whether my EA implementation is just not as good as theirs.

light-weaver:
Nevermind, I did some quick tests. Here's the reference front (red, generated by the authors) vs my attempt (blue).
image

The minima of f_1 is easy to calculate. f_1 = x_1 + 120 x_2. So minima of f_1 should occur at the lower bound of x_1 (0.5) and x_2 (4). Hence, min f_1 should be 480.5, which is exactly what my attempt recieves. How they arrived at this front, I do not know.

I went down a rabbit hole for this specific problem formulation. It seems a few different papers have played a game of telephone with this problem. At some point, even the meaning of x_1 and x_2 were swapped. Life has no meaning.

PR #194, opened 2024-11-22
Simulator and surrogate support documentation

Matskuu:
- Documentation for the evaluator for simulator and surrogate based problems and an example on how to use it
- Documentation for simulator files and an example of a simulator file (this file is meant for explaining things needed for simulator and surrogate based problems and so if there is something to be said about surrogates, it can be added here)
- How to guides for defining simulator and surrogate based problems in DESDEO

### DIFF
--- a/docs/api/desdeo_problem.md
+++ b/docs/api/desdeo_problem.md
@@ -28,6 +28,11 @@
     options:
         heading_level: 3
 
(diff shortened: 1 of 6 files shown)


gialmisi (commented):
I glanced at the changed files and things seems to be in order.

PR #219, opened 2025-03-20
scalarization.py refactored to use optional ideal and nadir points

pStefanec:
scalarization.py refactored to use ideal and nadir points from arguments if possible, two new functions added to utils.py which return only corrected ideal or nadir points


pStefanec:
attempt to fix failing check was not successful

Paurikova (approved):

Issue #172, opened 2024-09-05
Implementing quality indicators in DESDEO 2.0


gialmisi:
# Update 12.6.2025:

As our research group has various and urgent needs for different indicators, a priority list of indicators is given below with the most urgent indicator to be implemented on top. The state of each indicator (waiting to be implemented 🔴, being implemented 🟡, or done 🟢), its name, a reference for the indicator, and possible additional information, are given below.

If you start implementing any indicator, please update its status in the list below and assign yourself to the issue. Leaving a comment notifying other of you intentions is also a good idea.

## Indicator list

- [x] 🟢 R-metric, [reference](https://ieeexplore.ieee.org/abstract/document/8049301), the indicator has different variants: R-HV , which uses Hypervolume, and R-IGD which uses Inverted Generational Distance
- [ ] 🟡 PHI, [reference](https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=10115283), the indicator can measure the learning and the decision phases of an interactive method separately, it is based on the Hypervolume
- [x]  🟢 IGD+, [reference](https://www.sciencedirect.com/science/article/pii/S0377221720309620?via%3Dihub), the indicator is also known as the modified IGD
- [x] 🟢 R2 metric, [reference](https://link.springer.com/chapter/10.1007/978-3-540-88908-3_14), described in Section 14.4.8 in the reference
- [ ] 🟡 Epsilon indicator, [reference](https://doi.org/10.1007/978-3-540-30217-9_84), old implementation in [desdeo-tools](https://github.com/industrial-optimization-group/desdeo-tools/blob/master/desdeo_tools/utilities/quality_indicator.py)

### Template for adding indicators

Use this template to add new items to the above list:

```
- [ ] 🔴🟡🟢, INDICATOR NAME, [reference](link-to-reference), OPTIONAL ADDITIONAL INFO
```

---

We are currently missing quality indicators in version 2.0 of DESDEO. At least a couple indicators need to be implemented, documented, and tested.

A paper describing what quality indicators are in the context of (interactive) multiobjective optimization can be found [here](https://jyx.jyu.fi/handle/123456789/95813)

How indicators have been implemented in the older version of DESDEO, can be found here: https://github.com/industrial-optimization-group/desdeo-tools/blob/master/desdeo_tools/utilities/quality_indicator.py

[...]

PR #220, opened 2025-03-23
R-metric indicator

Paurikova:
Implementation of R-metric indicator.


light-weaver:
See this comment: https://github.com/industrial-optimization-group/DESDEO/issues/172#issuecomment-2746750042

PR #221, opened 2025-03-23
IGD+ indicator

Paurikova:
Implementation of the IGD+ indicator with a geometric average.

### DIFF
--- a/desdeo/tools/indicators_unary.py
+++ b/desdeo/tools/indicators_unary.py
@@ -181,35 +181,42 @@ def distance_indicators_batch(
         inds[set_name] = distance_indicators(solution_sets[set_name], reference_set, p=p)
     return inds
 
+
 class IGDPlusIndicators(BaseModel):
-    """A container for IGD+ indicator."""
-
-    igd_plus: float = Field(description="The IGD+ indicator value.")
-    "The IGD+ indicator value."
+    """A container for the IGD+ distance-based indicator."""
+    igd_plus: float = Field(description="The modified inverted generational distance (IGD+) indicator value.")
 
 
-def igd_plus_indicators(solution_set: np.ndarray, reference_set: np.ndarray, p: float = 2.0) -> IGDPlusIndicators:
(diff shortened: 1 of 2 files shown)


desdeo/tools/indicators_unary.py
```diff
@@ -181,6 +181,49 @@ def distance_indicators_batch(
...
+        reference_set (np.ndarray): The reference Pareto front.
+        p (float, optional): The power of the Minkowski metric. Defaults to 2.0 (Euclidean distance).
+
+    Returns:
+        IGDPlusIndicators: A Pydantic class containing the IGD+ indicator value.
+    """
+    distance_matrix = cdist(solution_set, reference_set, metric="minkowski", p=p)
+    min_distances = np.min(distance_matrix, axis=0)
+    ref_size = reference_set.shape[0]
+    igd_plus_value = (np.prod(min_distances) ** (1 / ref_size))
```
Paurikova, line 205: @gialmisi I used geometric average, isn't better aritmetic average?

Paurikova:
@gialmisi I just realized that in DistanceIndicators, the IGD+ indicator is already implemented, so we don't need to implement it again. We can close this PR without merging, can't we?

desdeo/tools/indicators_unary.py
light-weaver: Use the standard formulation of the indicators as published.

light-weaver:
DistanceIndicators implements something called IGD_p. That is not the same as IGD_plus. See this paper for the difference.
https://www.sciencedirect.com/science/article/pii/S0377221720309620

3.4.1 IGD
3.4.2 AHD (contains IGD_p and GD_p)
3.4.3 IGD+.

From that paper, you can go to the original references for more details.

desdeo/tools/indicators_unary.py
light-weaver: Also, if you are implementing some indicator yourself rather than using an implementation from some other package, do some tests to make sure your implementation gives the same results as these other packages. See, e.g., pymoo, moocore, hvwfg, pygmo, deap packages.

PR #222, opened 2025-03-30
Phi indicator

pStefanec:
Migrating PHI indicator from desdeo-tools. 

PHI calculation maybe works (at least it does not fail and provides seemingly reasonable values).

PHIDecision fails because the arrays in the assessment calculation have different shapes... Did it work in the desdeo-tools?


gialmisi:
Since I am not familiar with the phi indicator, could @light-weaver or @giomara-larraga have a look at the code? Assuming you are more familiar with the indicator. If not, let me know as well.

pStefanec:
@light-weaver could you take a look at the Phi indicator? On the last meeting you mention I should remove the classes. Do you want me to keep the PhiDecision functionality or it's not necessary (it was in desdeo1). I will have time to work on this until tomorrow evening

PR #223, opened 2025-03-31
Mcwb test problem

Paurikova:
Adding MCWB_SolidRectangular problem to desdeo.


desdeo/problem/testproblems/mcwb_problem.py
```diff
@@ -0,0 +1,86 @@
...
+    Constraint,
+Constant,
+    ConstraintTypeEnum,
+    ExtraFunction,
+    Objective,
+    Problem,
+    Variable,
+    VariableTypeEnum,
+)
+
```
gialmisi, line 11: Please document the problem by providing a docstring. It would be ideal if the docstring would contain the definition of the problem. For example, see [here](https://github.com/industrial-optimization-group/DESDEO/blob/0badca10505fbcfed520cbf3ce7a430e3bc560fe/desdeo/problem/testproblems/re_problem.py#L13):

```
    r"""Defines the four bar truss design problem.

    The objective functions and constraints for the four bar truss design problem are defined as follows:

    \begin{align}
        &\min_{\mathbf{x}} & f_1(\mathbf{x}) & = L(2x_1 + \sqrt{2}x_2 + \sqrt{x_3} + x_4) \\
        &\min_{\mathbf{x}} & f_2(\mathbf{x}) & = \frac{FL}{E}\left(\frac{2}{x_1} + \frac{2\sqrt{2}}{x_2}
        - \frac{2\sqrt{2}}{x_3} + \frac{2}{x_4}\right) \\
        &\text{s.t.,}   & \frac{F}{\sigma} \leq x_1 & \leq 3\frac{F}{\sigma},\\
        & & \sqrt{2}\frac{F}{\sigma} \leq x_2 & \leq 3\frac{F}{\sigma},\\
        & & \sqrt{2}\frac{F}{\sigma} \leq x_3 & \leq 3\frac{F}{\sigma},\\
        & & \frac{F}{\sigma} \leq x_4 & \leq 3\frac{F}{\sigma},
    \end{align}

    where $x_1, x_4 \in [a, 3a]$, $x_2, x_3 \in [\sqrt{2}a, 3a]$, and $a = F/\sigma$.
    The parameters are defined as $F = 10$ $kN$, $E = 2e^5$ $kN/cm^2$, $L = 200$ $cm$, and $\sigma = 10$ $kN/cm^2$.

    References:
        Cheng, F. Y., & Li, X. S. (1999). Generalized center method for multiobjective engineering optimization.
            Engineering Optimization, 31(5), 641-661.

        Tanabe, R. & Ishibuchi, H. (2020). An easy-to-use real-world multi-objective
            optimization problem suite. Applied soft computing, 89, 106078.
            https://doi.org/10.1016/j.asoc.2020.106078.

        https://github.com/ryojitanabe/reproblems/blob/master/reproblem_python_ver/reproblem.py

    Args:
        f (float, optional): Force (kN). Defaults to 10.0.
        sigma (float. optional): Stress (kN/cm^2). Defaults to 10.0.
        e (float, optional): Young modulus? (kN/cm^2). Defaults to 2.0 * 1e5.
        l (float, optional): Length (cm). Defaults to 200.0.

    Returns:
        Problem: an instance of the four bar truss design problem.
    """
```

gialmisi (changes requested):
This seems to be otherwise fine. I ask that the function is documented, like in the provided example in my comment.

gialmisi:
This the bug your experienced was due to a bug in how the JSON parser worked. This has now been fixed. It is a small wonder we did not notice this earlier, or that it has not caused problems, because it was quite a serious bug. I am happy it was discovered!

In the latest commit 8a8cf4 to desdeo2, this bug is fixed. If you sync your branch with the upstream, you should be able to evaluate the problem. I hope this helps!

Paurikova:
Closed based on PR https://github.com/industrial-optimization-group/DESDEO/pull/240

PR #240, opened 2025-05-01
Mcwb test problem

Paurikova:
Added mcwb problems from https://www.mdolab.net/Ray/Research-Data/mcwb_problem.zip class mcwb_concepts.py.


Paurikova:
@gialmisi 
Please take a look at my implementation to verify if the functions, objectives, etc., are correctly implemented. I wrote the test, but based on the original implementation, I was unable to check the values. Now, I am asserting values only based on the results from my implementation.

PR #241, opened 2025-05-01
documentation should use mostly python notebooks from now

pStefanec:
All markdown files replaced by python notebooks. 
No references to desdeo2 found. For TODOs check the issue description.

**Issues with the mathematical expressions found when checking the documentation locally (in the notebooks itself were all expressions displayed correctly)**


gialmisi (changes requested):
The notebook `problem_format.ipynb` needs to be addressed. It used templates, such as 

```python
{{ get_constant_info()[0] }}
```

that call the functions defined in `docs/explanations/macros.py`. If you look into the macro file, you will see what needs to be called to generate the needed output. If you[ look at the online documentation](https://desdeo.readthedocs.io/en/desdeo2/explanation/problem_format/), you see what the output should be. The end result in using the notebook does not need to be a carbon copy of the previous version, but it should at least display the schemas and example jsons that have been shown in the current version of the docs.

pStefanec:
@gialmisi I tried to fix the generated parts, the problem_format.ipynb has literally the same content as the original markdown file but it does not work for some reason... I guess some changes to the mkdocs.yml configuration file could help??? It is also highly possible that I overlooked some important detail.

gialmisi:
Thanks @pStefanec !

To address your questions, see my PR to your branch. I added an example on how to handle the issue with the blocks with the {{ ... }} format, and how to handle admonitions (markdown blocks that start with '!!! Note' etc..". Does this help?

pStefanec:
@gialmisi I checked the PR and I was just wondering if it would make more sense to leave the documentations which use macros heavily (like problem_format.ipynb) as markdown files if we are not able to make the whole process fully automatic? If not I can apply  the changes you suggested.

PR #243, opened 2025-05-12
Added ADM1, BLX-alpha crossover, and Uniform mutation

giomara-larraga:
I added the ADM1 to desdeo2, including a python notebook showing how to use it.

I also added BLX-alpha crossover, and Uniform mutation


gialmisi:
Can you check the `master` branch the the blend alpha and uniform crossover implemented there? If these can be used, can you update the PR to utilize these and target the current master branch instead of the desdeo2 branch?

PR #256, opened 2025-05-23
feat: 2 models and router for nimbus

StinaSoile:
- NIMBUSClassificationRequest
- NIMBUSClassificationState
- endpoint method/nimbus/solve

### DIFF
--- a/desdeo/api/app.py
+++ b/desdeo/api/app.py
@@ -9,6 +9,7 @@ from desdeo.api.routers import (
     reference_point_method,
     session,
     user_authentication,
+    nimbus,
 )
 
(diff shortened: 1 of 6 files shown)


gialmisi:
Good work! I had no comments, but I notice a print statement in the test case, which I removed. I suggest looking into using a debugger, which can avoid practically all uses of print statements when testing code.

I also realized that when using default values, e.g., for `solver_options`, then `None` is stored in the resulting state. Instead, this should contain the actual options used. However, this is a more general issue, and not specific to this PR. I will update this to the web-API mega issue.

PR #269, opened 2025-06-04
bug? fix for enautilus and scaling delta for group scalas

jpajasmaa:
New E-NAUTILUS has an error with intermediate_points type. I changed it to be  list[dict[str, float]] = Field(description="New intermediate points") and now it runs. If this is not intended change, revert this. 

For group scalarization functions, delta is now a dictionary for each objective function to be scaled before giving it as a parameter. Needed stuff for GNIMBUS, feel free to make more lasting fix.

### DIFF
--- a/desdeo/mcdm/enautilus.py
+++ b/desdeo/mcdm/enautilus.py
@@ -46,7 +46,7 @@ class ENautilusResult(BaseModel):
 
     current_iteration: int = Field(description="Number of the current iteration.")
     iterations_left: int = Field(description="Number of iterations left.")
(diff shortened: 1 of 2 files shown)


gialmisi:
`test_add_group_stom_sf_diff`, `test_add_group_guess_sf_diff`, and `test_add_group_asf_diff` are now failing due to `delta` being `None`. Is there a reason we should not use a default value (other than none) in case delta is not supplied? `delta=None` is not a good way to do this, since functions should not fail if executed with some arguments having their default values. If there is not good default for delta, then we should have delta be a required argument instead.

PR #275, opened 2025-06-09
[Do Not Accept] Benchmarks from external suites

light-weaver:
DESDEO currently does not have many benchmarking problems. Other packages do. DESDEO can technically call them by treating them as "simulators". This is very slow as DESDEO recreates subprocesses for each new evaluation.

I implemented another way to support simulators (in a very hacky, untested way). The simulator is expected to be accessed through a RestAPI. This seems much faster than the old way (though still very slow compared to native pymoo or DESDEO polars implementations).

### DIFF
--- a/BenchExp/experiment.ipynb
+++ b/BenchExp/experiment.ipynb
@@ -1,0 +1,338 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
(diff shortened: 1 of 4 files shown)


gialmisi:
Is this still relevant to be kept open?

PR #283, opened 2025-06-13
Update the hypervolume indicator to utilize moocore instad of pymoo

laitozz:
- I [compared](https://github.com/laitozz/DESDEO/commit/e70a862) the moocore and pymoo implementations and they seem to be equal down to rounding error.


light-weaver:
Looks good to me. I'd preferably use the method-based implementation (https://multi-objective.github.io/moocore/python/reference/generated/moocore.hypervolume.html#moocore.hypervolume) instead of the OO implementation, but I don't think this is a big deal in the grand scheme of things.

PR #286, opened 2025-06-15
Saving solutions to archive

StinaSoile:
- rename and restructure archive models
- create endpoint for nimbus for saving solutions
- create function in utils for both saving solutions to users archive and creating stateDB for it
- create models needed in saving
- make tests for the new endpoint and models

### DIFF
--- a/desdeo/api/models/__init__.py
+++ b/desdeo/api/models/__init__.py
@@ -1,8 +1,8 @@
 """Model exports."""
 
 __all__ = [
-    "ArchiveEntryBase",
-    "ArchiveEntryDB",
+    "UserSavedSolutionBase",
+    "UserSavedSolutionDB",
     "Bounds",
     "ConstantDB",
     "ConstraintDB",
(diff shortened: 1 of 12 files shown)


StinaSoile:
What still bothers me:
- all the solutions that are saved in the same request, will have the same method, since they all refer to the same state. I am not sure if the solutions saved at once should somehow need to be able to have different methods with each other. Of course they can refer to different methods if they are saved separately and different states are created in saving them.
- why does the UserSavedSolutionDB model accept so many None values? I wonder if state, user and problem should never be None.

gialmisi (changes requested):
This looks good, but  check my comment about `UserSavedSolverResults` and in `models/state.py`. Just minor changes.

gialmisi:
> What still bothers me:
> 
>     * all the solutions that are saved in the same request, will have the same method, since they all refer to the same state. I am not sure if the solutions saved at once should somehow need to be able to have different methods with each other. Of course they can refer to different methods if they are saved separately and different states are created in saving them.

I highly doubt that we would need to differentiate solutions to origin from different methods in the same request. So this is fine.

> 
>     * why does the UserSavedSolutionDB model accept so many None values? I wonder if state, user and problem should never be None.

This is just how SQLModel works. Some fields need to be able to be `None`, like `id`, so that we can create the model before adding it to the database where `id` will be assigned a value automatically. Being `None` does not mean the field is nullable (i.e., being `None` or NULL in the database itself). This is explained [here](https://sqlmodel.tiangolo.com/tutorial/automatic-id-none-refresh/) for `id`, but the reason is pretty much the same for other non-nullable `None` fields.

PR #288, opened 2025-06-16
Add epsilon indicator

laitozz:
- The implementation is taken from desdeo-tools
- Added some preliminary tests for the function.
    - The tests could definitely be improved but I couldn't figure out a better way to test the indicator.


light-weaver:
Merged. However, the implemented feature isn't _really_ the indicator, but rather just a component of it. The component (instead of the indicator) is an important part of IBEA so it is needed. I will rename this to "epsilon_component". We can use the indicator from moocore.

Issue #287, opened 2025-06-16
Problem schema: non-existent fields are accepted


gialmisi:
Currently, when creating a `Problem` class instance, or any instance of a class defined in the schema in `desdeo/problem/schema.py`, it is possible to include named arguments that do not match any existing field. E.g.,

```python
variable = Variable(..., lower_bound=0.5, ...)
```

where in the above example, the `lower_bound` argument is not defined (it should be `lowerbound`), which can lead to a lot of confusion. 

The needed change is to make the Pydantic models stricter in what fields they accept. Setting `extra` to `forbid` in the model configuration should do the trick: https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra

This change can lead to errors in the tests, which should also be checked and fixed, if possible.

PR #290, opened 2025-06-18
Forbid assigning values to undefined fields in pydantic classes

laitozz:
Also fixes the resulting errors. 
* Converting databases to pydantic classes now requires excluding the id fields. (which is bad UX?)
* Converting json to pydantic classes requires the by_name argument
    * FastAPI does not(?) support changing the default type coercion behaviour, so a helper function is needed, which is quite inelegant. Another option would be to override the model_validate function to use names by default.


laitozz:
#287

Issue #293, opened 2025-06-23
Collecting all available solvers and their default options into a dict


gialmisi:
To be able to communicate through the web-API what solvers are available in DESDEO, these need to be collected into a single dictionary, where the key is the name of the solver, which is followed by its class and settings object. I.e., for Bonmin, this would look like:

```python
available_solvers = {"Bonmin": {"constuctor": Bonmin, "options": BonminOptions}}
```

If some solvers have no Pydantic model for their options, please mention this in a reply to this thread.

__Additionally__: There are many default options, again, for Bonmin, there is `_default_bonmin_options = BonminOptions()`, which is fine. But there are also cases like with cbc:

```python
_default_cbc_options = CbcOptions(
    sec=600,
    threads=4,
    logLevel=2,
    maxSolutions=10,
    maxNodes=1000,
    ratioGap=0.01,
    absoluteGap=1.0,
    solve="branchAndCut",
    presolve=2,
    feasibilityTolerance=1e-6,
    integerTolerance=1e-5,
)
"""Defines CBC options with default values."""
```

This should be just `_default_cbc_options = CbcOptions()`. In other words, if using any defaults, they should be come from the Pydantic model (`CbcOptions`). These cases should be refactored.

All the solvers should be checked and these changes should be applied to them, if possible.

laitozz:
I opened a PR for passing options to the gurobi solvers. The initializer for ProximalSolver already takes a dict kwargs that is unused and defaults to None. Should I also add an empty model for that?

PR #295, opened 2025-06-24
Add solver option classes to the available_solvers dictionary

laitozz:
* Update the dictionary accesses accordingly. 
* Move the default values for the cbc solver to the pydantic model instead of _default_cbc_options.


laitozz:
#293

PR #298, opened 2025-06-26
Validate problem schemas by name. 

laitozz:
* Improves #290. 
* Removes the need to pass by_name to every validate call, which means that we don't need a helper function for parsing json into problem classes, which should be more elegant.


juropo:
I don't think validate_by_name should cause any serious problems, but this pull request is assuming that the web ui would call add_problem with a valid Problem python class as a parameter. I don't think that's going to happen.

PR #301, opened 2025-07-03
Metadata fetch

viljokass:
ProblemInfo and ProblemInfoSmall now return a listing of all metadata associated with the corresponding problem.

A separate endpoint is then used to fetch the metadata. Metadata requests include the corresponding problem's ID and the requested metadata type as a string that matches with the metadata_type field of different metadata.


viljokass:
@juropo fixed

PR #302, opened 2025-07-03
fixes to desdeo.tools.utils.py

viljokass:
Came across these bugs while testing with data from RPM.
* Line 71: In available_solvers dict, changed key "guropipy" to "gurobipy".
* Line 169: Noticed that despite the problem having TensorVariables, execution of the program didn't enter the if-block.


juropo:
Yeah, this seems good.

PR #306, opened 2025-07-08
Inputs for preferences and numSolutions

StinaSoile:
- In API, I added previous_preference to NIMBUSClassificationState, since it is needed in the future for initializing preferences

- preference list, numSolutions and its min and max values come as optional props from nimbus page to AppSidebar
- I added input for preference and name, possible unit and min/max info for avery objective, like it was in old UI
- I copied the classification value logic from old UI, too, so that the texts "change freely", "Worsen until" etc are also visible for every objective.
- inputs for preference and number of solutions are clamped to min or max, if user tries to write invalid number. For preference, min and max are ideal and nadir.
- preference and numSolutions are saved in states in AppSidebar, so that they stay in sync everywhere
- they are also in states in nimbus +page, and onChange-function updates them

### DIFF
--- a/webui/src/lib/components/custom/preferences-bar/preferences-sidebar.svelte
+++ b/webui/src/lib/components/custom/preferences-bar/preferences-sidebar.svelte
@@ -2,17 +2,27 @@
 	import * as Sidebar from '$lib/components/ui/sidebar/index.js';
 	import PreferenceSwitcher from './preference-switcher.svelte';
 	import { writable } from 'svelte/store';
+	import { Button } from '$lib/components/ui/button/index.js';
 	import type { components } from '$lib/api/client-types';
-	import { HorizontalBar } from '$lib/components/visualizations/horizontal-bar';
+	import {
+		HorizontalBar,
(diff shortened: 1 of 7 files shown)


StinaSoile:
Problems and todos:
- latest merge caused a problem, @giomara-larraga knows how to fix it

- I think it is weird to have the classification value creation in a component that is general and not for nimbus. But I dont know what to do with it.
- preferences are not actually initialized based on NIMBUSClassificationState.previous_preferences: that whole state is not fetched anywhere
- previous preferences are not shown for every objective, like in the old UI, because I dont have that information right now
- ideal and nadir can be null or undefined according to typescript validation, so I needed to do type guards everywhere. But they said in the meeting that they should not be null or undefined at this point, so something is wrong somewhere else
- came up in the meeting: values have the default amount of significant numbers, but that should be defined in problem metadata and fetched and if it is not there (it is not), then use default.
- came up in meeting: in props, dict of text may be needed, where there is the info of method and language
- something else probably came up, too, I just couldn't write everything down

PR #307, opened 2025-07-11
feat: minimal nimbus initialization endpoint

StinaSoile:
I created an endpoint for initializing nimbus. 
This returns a state with solver_results created with generate_starting_point function.
I created NIMBUSInitializationRequest and -State for this.

I didn't find out all that it should do, but as far as I understood, the start_results are the main thing and what I think I need when I continue figuring out GUI.

"It should also probably try to figure out if this is the first time of loading up the specific problem or not, and act accordingly", but for now it doesn't, because I don't know what this includes. Maybe calculate ideal and nadir if they don't exist? Something else?
I can come back to this later if needed.

### DIFF
--- a/desdeo/api/models/__init__.py
+++ b/desdeo/api/models/__init__.py
@@ -18,6 +18,8 @@ __all__ = [
     "NIMBUSBaseState",
     "NIMBUSClassificationRequest",
     "NIMBUSClassificationState",
+    "NIMBUSInitializationRequest",
+    "NIMBUSInitializationState",
     "NIMBUSSaveRequest",
     "NIMBUSSaveState",
     "ObjectiveDB",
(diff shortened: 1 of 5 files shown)


StinaSoile:
I changed this so that it first tries to find previous NIMBUS states.
Now this tries to find the latest NIMBUSClassificationState or NIMBUSInitializationState, and return that,
and if that does not exist, it creates a new NIMBUSInitializationState.

 - I understood that the history logic should be such that initialization will create a new branch to the history tree from the point of the latest nimbus state, or from the beginning if there are no previous nimbus states. In the old version, initialization searched for all solutions with a nimbus method and chose either the one marked as current, or just the first one. But with the new state history, it is a tree, so I'd guess the latest is always the right one? Or should there be a way to choose which one of found states will be returned?
 
 - Are all three, NIMBUSClassificationState, NIMBUSInitializationState and NIMBUSSaveState, valid states that initialization should find? Code is simpler to fix from ignoring one to accepting all, and that's why it ignores SaveState now.
 
 - Can initialization request ever have a parent_state_id in request? I can easily add it if needed, I just understand that initialization either continues from existing or creates completely new branch from beginning and thus it cant have parent_state given in request.

PR #312, opened 2025-07-24
added test forest problem with metadata to db

viljokass:
Added:
* utopia_db_init.py, which initializes a database that has utopia problem and corresponding database
* the data and some tools to create the database entries from the data

@StinaSoile @light-weaver

### DIFF
--- a/tests/data/forestProblemWithMetadata/utopia_db_init.py
+++ b/tests/data/forestProblemWithMetadata/utopia_db_init.py
@@ -1,6 +1,7 @@
 """This module initializes the database."""
 
 import warnings
+import os
 import json
 
 from pathlib import Path
@@ -40,7 +41,8 @@ if __name__ == "__main__":
             session.commit()
             session.refresh(user_analyst)
 
-            path = Path(__file__)
+            current_path = Path(__file__)
+            path = current_path
             while not str(path).endswith("/DESDEO"):
                 path = path.parent
             path = path / "tests/data/forestProblemWithMetadata/"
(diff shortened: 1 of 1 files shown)


StinaSoile:
When I try to use this in the UI in the nimbus page, initialization fails trying to run **generate_starting_point()**. Seems like there is NoneType somewhere in bounds where it should not be. I have no idea what this is about.

End of error message that might be useful:

  File "/home/stsosupa/Work/DESDEO/desdeo/api/routers/nimbus.py", line 181, in initialize
    start_result = generate_starting_point(
                   ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stsosupa/Work/DESDEO/desdeo/mcdm/nimbus.py", line 415, in generate_starting_point
    asf_solver = init_solver(problem_w_asf, _solver_options) if _solver_options else init_solver(problem_w_asf)
                                                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stsosupa/Work/DESDEO/desdeo/tools/pyomo_solver_interfaces.py", line 432, in __init__
    self.evaluator = PyomoEvaluator(problem)
                     ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stsosupa/Work/DESDEO/desdeo/problem/pyomo_evaluator.py", line 53, in __init__
    model = self.init_variables(problem, model)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/stsosupa/Work/DESDEO/desdeo/problem/pyomo_evaluator.py", line 175, in init_variables
    else (var.lowerbound + var.upperbound) / 2
          ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for +: 'float' and 'NoneType'

viljokass:
I see what the problem is. For now, try using gurobipy evaluator, it has worked in the past.

juropo:
If this is for testing purposes, it would be better to have the db_init somewhere else than under api. Also, the data files could go to the pre-existing folder containing test data i.e. tests/data

In the long term, we probably also don't want a folder called utopia_stuff, but it is what it is.

PR #315, opened 2025-08-01
Group Decision Making implementation

viljokass:
# Group Decision Making documentation
This is a documentation on how the group decision making is implemented, how it can be modified to generalize different preference information, etc. I began expanding the WebSocket example from FastAPI documentation and it grew into this.

Tool that have been used in testing:
* websocketking.com: allows creating websockets and sending data through them.

Relevant and modified files in DESDEO:
* desdeo/api/app.py: added the gdm router
* desdeo/api/models/\_\_init\_\_.py: added the gdm models
* desdeo/api/models/gdm.py: added group decision making models
* desdeo/api/routers/gdm.py: added group decision making functionality
* desdeo/api/tests/test\_models.py: testing
* desdeo/api/tests/test\_router.py: testing
* pyproject.toml: added websockets as dependency

## Notes on the implementation
To start, in the routers/gdm.py file we can see a number of endpoints for group manipulation. We can create and delete groups, add and remove users from them, initialize the group's problem (especially in the case of nimbus), get group info and fetch the results of the latest iteration of the group. These work like any other HTTP endpoint accross the DESDEO API. There is one new type of endpoint, and that is the websocket endpoint.

### The models
The models in models/gdm.py consist of two main table models: Group table model and GroupIteration table model. There are some response and request models and a few data models to be included within the GroupIteration. The Group model holds data on the Group: id, name, owner's id, list of group member ids and the id of the problem that the group has. It also has a reference to the latest GroupIteration known as head. The GroupIteration model handles the data within the iterations: the preferences of that iteration, the results of that iteration, whether users have been notified, and references to the group, iteration's parent and its child.

The results/preferences are stored in separate structure, which can be derived into many different forms. As of now, there's only the NIMBUS preference/results type and voting preference/result type. The structure contains a "method" field. We can use that to determine the behaviour of the preference setting and optimization process. When optimization happens, a new GroupIteration with the next phase's type is created, to be filled with preferences from the users.

As an example, suppose we have a GDM method that combines NIMBUS and voting on the NIMBUS results.
Visually, the Group/GroupIteration relationship would look like this.

                             +-------------+
                             |             |
     _______         ________v_______      |     ________________
    | Group | +---->| GroupIteration |  +--+--->| GroupIteration |  +-----> etc.
    |_______| |     |----------------|  |  |    |----------------|  |
    | head ---+     | parent -----------+  |    | parent -----------+
    |_______|       | child  -----------+  |    | child  -----------+
                    | VotingPrefRes  |  |  |    | NIMBUSPrefRes  |  |
                    |________________|  |  |    |________________|  |
                                        |  |                        |
                            None <------+  +-------------+----------+

Here, we would be expecting voting preferences from the users.

### ManagerManager and GroupManager
When the server is started, a singleton called ManagerManager that manages connection/preference/optimization managers is spawned. Its responsibilities are to create and delete these managers. If a valid connection, with authorized user connects, ManagerManager (NOTE: check notes on ManagerManager; it doesn't do that just yet, just returns nimbus manager) checks whether a corresponding group/method already has a group manager and returns it. If there is no such group manager, it spawns it. Upon disconnection of a user, ManagerManager checks whether a group manager has any remaining connections. If there are none, the singleton deletes this manager.

A GroupManager handles the needs of a single group. These needs are handling connections, database access, broadcasting messages, preference updates and optimization. Lets's go over its functionality function by function.

`
def __init__(self, group_id: int)
`
Naturally, this is the constructor of the object. It checks that the group corresponding to the id exists. Then it initializes a dictionary for the users' websockets.

`
async def send_message(self, message: str, websocket: WebSocket)
`
Just sends single message through the specified websocket.

`
async def connect(self, user_id: int, websocket: WebSocket)
`
Attaches the websocket in the GroupManager's websocket dictionary. Also, if the previous iteration has any notifications that aren't sent, send notifications that the results are to be fetched.

`
async def disconnect(self, user_id: int, websocket: WebSocket)
`
Deattaches the websocket object from the GroupManager's websocket dictionary.

`
async def broadcast(self, message: str)
`
Sends a message to all connected websockets.

`
async def notify(self, user_ids: list[int], message: str):
`
Notifies all users that are connected to the Manager. Returns a dict of who is notified and who is not.

Those would be basic handling of connections and such. However, the difference between managers come from the "optimize" function and what happens within.

`
async def optimize(self, user_id: int, data: ReferencePoint)
`
This function is quite free form. As an example I have implemented a simple NIMBUS thing. Feel free to expand on that.

We check the group and fetch its latest iteration. We check the preference/result type of the iteration and using that we diverge the execution. Currently only "nimbus" type has its execution implemented. Next, we validate the data from the websocket and update the iteration's preferences in the database. If all preferences are in, we begin optimization. If optimization succeeds, we update the results of the iteration in the database. Then we create a new iteration (in this case with NIMBUS preference/result type, but if we want to implement some voting system, we could make voting preference/results), and update the parent/child relationships of the new and the current iterations and update the group's head iteration to the new iteration. The head iteration is then to be filled by preferences from the users.

Suppose you wanted to create a RPM-GDM. You'd derive a RPMManager from GroupManager and implement an "optimize" function (or whatever you want it called, really) that handles:
* Preference validation
* Preference updates to the database
* Optimization and result insertion into database
* Updating the iterations:
    * Current iteration -> parent of the newest iteration and relationships between those
    * Group head iteration -> newest iteration

In addition you'd need a RPMPreferenceResult data object derived from BasePreferenceResult object that contains the preference information and the corresponding results.

### Database connections
The system can handle a small number of database connections simultaneously. However, I was unable to completely eliminate the database connection pool from drying up with maximum connection count of 5 + 10. I've tried to make sure that each GroupManager has only one active connection at a time, however, that doesn't seem to have solved to problem or my skills are lacking. It should be able to handle light loads, but with stress testing by bombarding with user sending data through the websocket, the pool dried up. 

### Relation to states
The current implementation doesn't yet use the state system that the surrounding framework uses. This, however, shouldn't be too difficult to implement; a state id to the GroupIteration and the creation of these states (perhaps using the existing HTTP endpoints) would probably do the trick. The user required in these calls could possibly be the group owner.

### WebSocket endpoint
The URL of the websocket endpoint is
`
ws://[HOST]:[PORT]/gdm/ws?group_id=[GROUP_ID]&method=[METHOD]&token=[TOKEN],
`
where group id is the user's group's id, method is the method we're gonna use (like the NIMBUS thing described above; currently only responds to "nimbus") and the token is the access token from the API /login endpoint. From the endpoint we authenticate the user and receive data continuously. When data is received, call the manager's "optimize" function with the data and user's ID. In the case of this NIMBUS thingy, valid data sent over the socket would be for example:
`
{"aspiration_levels": {"f_1": 0.4, "f_2": 0.7, "f_3": 0.1}}
`
But in general the type and format of the preference is dependent on the method.

In testing I've been using websocketking.com. It allows creating websockets and saving payloads and such. Upon entering the site it should show a little tutorial and it explains itself pretty well. Just keep the above URL and payload in mind.

In order to start working with the websockets, one must:
* create group
* initialize the group (in the case on this implemented nimbus method)
* add all wanted users to the group
* log in to those users and get the access token to be inserted into the URL query


gialmisi:
Also, @viljokass , add the content of the OP in the docs as well. It explains nicely how things work, and can be very useful for others.

PR #319, opened 2025-08-12
Nimbus UI: new endpoints,  intermediate, finish and save, graph, map

StinaSoile:
- BACKEND: Vili had started the changes to endpoints and db in his own [branch](https://github.com/viljokass/DESDEO/tree/nimbus_fix), so I continued that in my branch to a point where I can use them in UI.
- MAP: Added map fetching if that exists. The map is exactly the same as in the previous UI. I will make it more beautiful. There is a pull request #312 from Vili that has a forest problem that I used to test the map. It can be used to check how the map looks like, but I don't think it completely works.
- INTERMEDIATE: User can choose to solve intermediate solutions on top of the sidebar (ugly, causes the problem in page height). This changes table and graph behaviour so that user can choose two solutions from either of them
- SAVE: saving happens in table, giving optional name to a solution and saving them one by one. There is a confirmation pop up that opens if user has unsaved changes to name. It behaves a little weird for now, I need to change it later. For example in intermediate mode it does not act nicely with the possibility to choose two solutions.
- GRAPH: graph and table are in sync, graph shows previous preference and previous solution in addition to current ones.
- DISPLAY ACCURACY is checked, but it does not exist in problem metadata so it ends up using fallback. I think this could be different for all objectives and I don't know if this should be the significant numbers (toPrecision()) or number of decimals (toFixed()). I thought it should be significant numbers but existing code used toFixed(). But I think this is easy to change later.


StinaSoile:
Changing the endpoints broke the EMO saving:
We ended up changing the saving function so that instead of the whole solution, it uses the address of the solution (it needs a state and index of solution in there).
Now the EMO is using the same saving function and of course it does not work anymore, since it asks for different type of solution.
Vili pointed this out already before his vacation, but I just forgot about it.

PR #323, opened 2025-08-25
Test case for using nimbus and intermediate solutions with forest problem

StinaSoile:
- I did the test_forest_nimbus_intermediate.py. It _kind of_  passes, so the problem might be in DB, then. In the general intermediate solutions route, I am using variables fetched from db as they are.
- What I mean by _kind of_  passing, is that  **tests complained about forest_problem.py**, tests being this and test_testproblems.py, that uses forest problem too. Complaint was **ValueError: setting an array element with a sequence**.
I let copilot "fix" forest_problem.py, so that tests would pass.  I didn't study what copilot did to the problem, but I kept the changes in case they are helpful. I dont understand the forest_problem.py, but if the tests complain about it, my thought is that maybe there is a problem.
- I also noticed that test_testproblems.py had a wrong address for forest problem, so I changed it.
- Since you were only going to pull this and use the test case, I merged the Vilis test forest problem too, so I can better try things.
- And yeah, the three other problems (dtlz2, simple knapsack, river pollution problem) all work solving intermediate solutions in UI.

### DIFF
--- a/desdeo/problem/testproblems/forest_problem.py
+++ b/desdeo/problem/testproblems/forest_problem.py
@@ -62,13 +62,7 @@ def forest_problem(simulation_results: str, treatment_key: str, holding: int = 1
     for i in range(np.shape(v_array)[0]):
         for j in range(np.shape(v_array)[1]):
             if (unique_units[i], j) in rows_by_key:
-                # Fix: Extract the single value from the sequence
-                try:
-                    # If it's a single value in a list/tuple
-                    v_array[i][j] = rows_by_key[(unique_units[i], j)][0]
-                except (ValueError, TypeError):
-                    # If it's a sequence itself (e.g., list of values)
-                    v_array[i][j] = rows_by_key[(unique_units[i], j)][0][0]
+                v_array[i][j] = rows_by_key[(unique_units[i], j)][0]
 
     # determine whether the results are to be compared to those from the rahti app (for testing purposes)
     # if compared, the stock values are calculated by substacting the value after 2025 period from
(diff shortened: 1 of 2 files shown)


gialmisi:
First, I notice that in the database, the problem is loaded from a file `problem = Problem.load_json(Path("/home/kilo/workspace/DESDEO/datasets/forestProblemWithMetadata/problem.json"))`, while in the test case it is defined using a test problem. These two problems where different. I decided to continue using the one loaded from a file since I was able to reproduce the error we got earlier about inferring a supertype when trying to evaluate the problem using the Polars evaluator.

While testing this, I noticed two things:

1. The solvers from the COIN-OR project all complain about the problem being either unbounded of infeasible. While the Gurobi solver is able to produce solutions, they are based on heuristics, which makes me suspect the problem formulation might not be correct, or the many unbounded variables cause problems. To my understanding, this should be a linear combinatorics problem, which should not require heuristics. Double check the problem definition (and which problem is being used) and see if any undefined bounds could be feasibly defined.
2. The Polars evaluator has some trouble in adding a scalar to a vector, when the scalar is defined as a constant. However, expressions like `5+X` where `X` is some arbitrary non-scalar tensor, work. Troubles arise when `5` is replaced by, for instance, a `Constant` or `Variable` `c`, which represents a scalar. This will lead Polars to try and reduce the a table in the polars evaluator with an expression with non-matching dimensions, which will lead to errors. Best course of action is to define the constants as part of the expression, i.e., instead of `c + X`, have `c + 5` explicitly. This bug can be fixed in the `json_parser.py` file, but I will not have time to check this. This is not trivial, since at the time of building the polars expression in `_polars_reduce`, we do not know anything about the size of the columns the expressions represent.

juropo:
I assume there is nothing in this pull request that we want to add and that hasn't been added already.

PR #330, opened 2025-08-29
API State refactor

gialmisi:
I have now gone through how we implement and use states in the API. There are some significant changes. Unfortunately, I did not have time to address the existing EMO stuff, so those need to be checked and updated. If @light-weaver and @viljokass you want to check and see how this fits into your work and then update this PR, that would be a good first step. After that, @StinaSoile and @giomara-larraga can pull the changes and see what kind of changes the frontend needs (though, the response and request models that the endpoints require and emit have changed minimally, expect for the initialization endpoint for NIMBUS.). I also suggest updating the OpenAPI clients on the webui (see the README in `webui`).

Still needs some polish, but the functionalities are now there. Here are some other notes to consider when integrating the changes:

- When defining a new method state, update `StateKind`, `KIND_TO_TABLE` and `SUBSTATE_TO_KIND` in `generic_states.py`. Add method specific states to `state.py`.
- If a state contain solutions that have been computed, it should implement the interface `ResultInterface`.
- Use the class method `StateDB.create()` when adding a new state to the database. __DO NOT__ use `StateDB()`.
- The previous `SolutionAddress` and friends, have seen big changes. See `SolutionReference` and `SavedSolutionReference` in `generic_states.py`. These are neat utilities where you just plug in a `StateDB` or `UserSavedSolutionDB`, respectively, and get a handy interface to access the state's contents. See the NIMBUS endpoints and tests for examples on how these work.
- `user_save_solutions` in `uitls/database.py` not really needed anymore. The EMO save request and state can follow the same logic as done in NIMBUS.


viljokass:
Looks good to me. I don't have anything to add for now as I'm working on GDM stuff at the backend which is still its separate branch. When these updates are in the main branch I'll try to connect the state and saving system to GNIMBUS.

light-weaver:
Looks good to me as well.

desdeo/api/routers/generic.py
```diff
@@ -45,6 +47,48 @@ def solve_intermediate(
...
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=(
+                    f"The index {solution_info.solution_index} is out of bounds for results with len={len(_var_values)}"
+                ),
+            ) from exc
+
+        try:
+            _obj_values = solution_state.state.result_objective_values
+            obj_values = _var_values[solution_info.solution_index]
```
StinaSoile, line 80: obj_values = _var_values[solution_info.solution_index]
should be 
obj_values = _obj_values[solution_info.solution_index]

StinaSoile:
What I noticed:
- I had made initialization endpoint so that it checks for previous states, and if they don't exist, it creates the initial state. 
Now it only creates an initial state and there is no separate endpoint for fetching the latest state.
I made get-or-initialize endpoint for my local branch, but it would be as easy to have endpoint only for getting the last state and then call initialization separately in UI if that does not exist. Idk. Does it matter.
Anyway, just one more endpoint needed.

- in EMOState and EMOSaveState there are these types list["EmoResults"] that caused an error that prevented generating the new client types. Not my problem, I dont care.

- In routes/general.py, there were a couple of typos in code, that broke stuff in UI. I left the comments.

Otherwise, I just basically renamed stuff in UI and all seems to work.

PR #343, opened 2025-09-18
Solver metadata

viljokass:
This allows attaching a preferred solver to a problem through metadata. There's also an endpoint for doing so. Also added the checking of this metadata to nimbus, gnimbus and rpm endpoints.

### DIFF
--- a/desdeo/api/app.py
+++ b/desdeo/api/app.py
@@ -32,6 +32,9 @@ app.include_router(nimbus.router)
 # app.include_router(EMO.router) #TODO: after EMO stuff works, put it to use again
 app.include_router(generic.router)
 app.include_router(utopia.router)
+app.include_router(gnimbus.router)
+app.include_router(gdm_base.router)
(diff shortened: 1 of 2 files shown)


gialmisi:
Seems fine. Had to modify some of the imports to avoid cyclical imports. When importing inside a module from other files, it is good to avoid importing from the said module. I.e., if we are working in the file `foo.py` in module `module`, and we want to import the function `tee` from another file in the same module `bar.py`, we should write:

```
# in module/foo.py

from module.bar import tee

...
```

instead of

```
# in module/foo.py

from module import tee

...
```

This is because in the `__init__.py` of `module` we can have

```
# module/__init__.py

from .foo import stuff
from .bar import tee, other_stuff
```

Now, if we used something like `from module import tee` in `foo.py`, we will try to import this from the `__init__.py`, which will then try to import things from `foo.py` and `bar.py`, which at this point are only partially initialized, leading to circular imports.

PR #349, opened 2025-09-24
Nimbus UI: waiting symbol, shortcut link to nimbus, errors

StinaSoile:
I made the spinner notification component, looks like in the picture:
image

Added a link to nimbus in dashboard.
- it magically knew to choose problem 1 and open it in nimbus.
- The problem can be whatever the first one happens to be, but the text is talking about forest management. Isn't the text misleading? I guess it does not matter, at least right now?

Other, more irrelevant stuff:

Fixed errors, they were not shown properly or even improperly
- I removed unnecessary error handling that didn't even work
- made error notification component to actually show the errors

I also ended up rearranging and Prettifying stuff in +server.ts and handlers.ts and helper-functions.ts, causing stupid amount of changed code, 
and removed a couple of files that were my old mess from moving stuff around. Sorry.


juropo:
I will load this up on Rahti tomorrow, and see if it works as it should.

juropo:
Seems to work great

PR #351, opened 2025-09-24
Now the configs should actually work

juropo:


juropo:
The change history looks awful because I forgot to sync my fork after changes yesterday, but there isn't actually this much stuff

PR #361, opened 2025-10-07
Gnimbus UI

StinaSoile:
Latest changes of gnimbus ui. 
Still missing the end view that makes sense, and might include stupidity.


juropo (commented):
Reading through all the UI changes seems like too much, but I didn't notice any big problems with the workings of the UI. 

No problems with the changes to the core stuff.

PR #381, opened 2025-11-14
EMO changes.

gialmisi:
- Fix for fast non-dominated sorting and test cases that do not pass with the current upstream version (but pass here).


desdeo/tools/non_dominated_sorting.py
light-weaver: `not fronts[i].any()` does not do anything anymore. You can remove it (and make things clearer). Otherwise this can be merged.

PR #390, opened 2025-12-18
New shceme to interface to external problems

gialmisi:
- Introduce a provider/locator architecture for external (non-DESDEO) problems.
- Add URI-based provider resolution with late binding at runtime.
- Implement ProviderRegistry and ProviderResolver with caching support.
- Allow providers to be registered dynamically during runtime.
- Add a Pymoo provider with cached problem construction and evaluation.
- Keep Problem objects fully serializable (no embedded callables).
- Support batch evaluation.

### DIFF
--- a/desdeo/problem/external/__init__.py
+++ b/desdeo/problem/external/__init__.py
@@ -1,10 +1,18 @@
 """Export of the external module."""
 
 from .core import ProviderParams
-from .pymoo_provider import PymooProvider
+from .pymoo_provider import PymooProblemParams, PymooProvider, create_pymoo_problem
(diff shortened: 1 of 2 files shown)


tests/test_external_problems.py
```diff
@@ -0,0 +1,116 @@
...
+    xs_3d = {s: [0.5, 0.5, 0.5] for s in var_symbols}
+
+    res_3d = evaluator.evaluate(xs_3d)
+
+    assert (obj_symbols | obj_symbols_min).issubset(res_3d.columns)
+    assert res_3d.height == 3
+
+
+@pytest.mark.external
+def test_w_emo():
```
light-weaver: common emo w 💪

PR #398, opened 2026-01-15
Endpoint for getting available solvers and its unit test.

PtrBednarik:


desdeo/api/routers/problem.py
```diff
@@ -235,6 +235,13 @@ def get_metadata(
     return [metadata for metadata in problem_metadata.all_metadata if metadata.metadata_type == request.metadata_type]
 
 
+@router.get("/assign/solver", response_model=list[str])
+def get_available_solvers() -> list[str]:
+    """
+    Return the list of available solver names.
```
gialmisi: This should give you a yellow squiggly line, if you have `ruff` and its extension setup correctly in vscode (should be just a matter of installing the ruff extension.) The first line of the docstring should start on line 240.

gialmisi: Otherwise the changes to the web-API code seem fine!

gialmisi (changes requested):
The PR should not include the changes to `.gitignore`.

Issue #406, opened 2026-01-21
Web-API: new endpoints for managing representative solution sets for problems


gialmisi:
# Description

We are currently missing endpoints for managing representative solution sets of a problem in the database as metadata.

# Requested changes

- [x] A new POST endpoint "/add_representative_solution_set" in `routers/problem.py` to add a new solution set as metadata to a problem in the database (`RepresentativeNonDominatedSolutions`). This endpoint expects a representative solution set as a JSON file contained in a new custom request type.
- [x] A new GET endpoint "all_representative_solution_sets/{problem_id}", which returns all sets that exist for a problem. The returned information should be meta-level, i.e., not the full data, but information such as its name, description, and ideal and nadir point values. This will also require a new type of response to be defined.
- [x] A new GET endpoint "representative_solution_set/{set_id}", which returns the full information on a requested set. This will also require a new type of response model.
- [x] A new DELETE endpoint "representative_solution_set/{set_id}", which deletes a set with a requested id, if it exists. Returning only an HTTP code is enough.
- [x] New endpoints and models should have at least basic tests defined as unit tests. Since our test suite is quite large, it is time we start writing unit tests in more specific files. So for this issue, all new tests should be defined in a new file under `api/tests`, such as `test_problem_metadata.py`. Existing tests do not need to be moved around.

# Things to note

The test case `test_problem_metadata` in `api/test/test_models.py` shows how the metadata can be handled for `RepresentativeNonDominatedSolutions`.

For the required new request and response models, the basic idea of these is documented in [the docs](https://desdeo.readthedocs.io/en/latest/explanation/database_models_and_schema/). It is worth reading, and feedback on it is also welcome! Existing models for responses and requests, should also be helpful references.

This issue should __not__ require for any new table (i.e., database) models to be created.

PR #418, opened 2026-02-03
Endpoints for managing representative solution sets for problems.

PtrBednarik:
https://github.com/industrial-optimization-group/DESDEO/issues/406

### DIFF
--- a/desdeo/api/routers/problem.py
+++ b/desdeo/api/routers/problem.py
@@ -329,13 +329,13 @@ def select_solver(
     response_model=RepresentativeSolutionSetInfo
 )
 def add_representative_solution_set(
-    payload: RepresentativeSolutionSetRequest,
-    context: Annotated[SessionContext, Depends(get_session_context_without_request)],
+    request: RepresentativeSolutionSetRequest,
+    context: Annotated[SessionContext, Depends(get_session_context)],
 ):
     """Add a new representative solution set as metadata to a problem.
 
     Args:
-        payload (RepresentativeSolutionSetRequest): The JSON body containing the
+        request (RepresentativeSolutionSetRequest): The JSON body containing the
             details of the representative solution set (name, description, solution data, ideal, nadir).
         context (SessionContext): The session context providing the current user and database session.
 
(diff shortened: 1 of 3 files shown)


desdeo/api/models/request_models.py
```diff
@@ -0,0 +1,12 @@
+"""Requests models."""
+from pydantic import BaseModel
+
+
+class RepresentativeSolutionSetRequest(BaseModel):
```
gialmisi: It would be very nice if we can have a base class that the table,request (and eventual response) models could inherit from. By table model I mean this https://github.com/industrial-optimization-group/DESDEO/blob/2b5d0bd7efa8dfc633864d63da609645c86b08a2/desdeo/api/models/problem.py#L243

The base model can very well have the same fields as this current model. This will require a small amount of refactoring. (feel free to move the current `RepresentativeNonDominatedSolutions` model from problem.py to here, if it makes sense to you).

desdeo/api/models/representative_solution.py
```diff
@@ -0,0 +1,32 @@
...
+class RepresentativeSolutionSetBase(SQLModel):
+    """Shared base model for representative solution sets."""
+
+    name: str
+    description: str | None = None
+    solution_data: dict[str, list[float]]
+    ideal: dict[str, float]
+    nadir: dict[str, float]
+
+class RepresentativeSolutionSetRequest(RepresentativeSolutionSetBase):
```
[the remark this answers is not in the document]
gialmisi: This is actually something we need to unify in the framework.

desdeo/api/routers/problem.py
```diff
@@ -318,3 +323,170 @@ def select_solver(
...
+        HTTPException: If problem not found or unauthorized user.
+
+    Returns:
+        dict: Confirmation message.
+    """
+    user = context.user
+    db_session: Session = context.db_session
+
+    # Fetch the problem
+    problem_db = db_session.get(ProblemDB, payload.problem_id)
```
gialmisi: I think there are now checks that would not be needed if we used `get_session_context` instead.

desdeo/api/routers/problem.py
```diff
@@ -318,3 +323,170 @@ def select_solver(
...
+        id=repr_metadata.id,
+        problem_id=problem_db.id,
+        name=repr_metadata.name,
+        description=repr_metadata.description,
+        ideal=repr_metadata.ideal,
+        nadir=repr_metadata.nadir,
+    )
+
+@router.get(
+    "/all_representative_solution_sets/{problem_id}",
```
gialmisi: Exactly, with endpoints operating on just id's, we should use query params. (just a comment)

gialmisi (changes requested):
Good. Just some minor retouches. (you really seem to love removing documentation... :D careful with that)

Issue #421, opened 2026-02-05
Web-API: purging singleton request models and using query params instead


gialmisi:
# Description

Currently, we have routers in the Web-API that use request models in their input, which only contain a single integer, usually an id of some kind. See for example:

https://github.com/industrial-optimization-group/DESDEO/blob/1100f19e91629d4645cb167ae514a5e047cbe425/desdeo/api/routers/problem.py#L88C1-L92C18

and the related model

https://github.com/industrial-optimization-group/DESDEO/blob/1100f19e91629d4645cb167ae514a5e047cbe425/desdeo/api/models/problem.py#L54-L57

Instead of having these models, we should just use query parameters, like here:

https://github.com/industrial-optimization-group/DESDEO/blob/1100f19e91629d4645cb167ae514a5e047cbe425/desdeo/api/routers/session.py#L48

# Requested changes

The routers in the Web-API should be checked for the described kind of endpoints, and modified to use query parameters instead. The unused request models can then be removed.

PR #433, opened 2026-02-23
Purging singleton request models and using query params instead

PtrBednarik:
https://github.com/industrial-optimization-group/DESDEO/issues/421


gialmisi (approved):
Good. I am happy to see that the net amount of documentation/comments has not decreased this time 😄

Issue #422, opened 2026-02-05
Web-API: implementing NAUTILUS Navigator endpoints


gialmisi:
# Description

The NAUTILUS Navigator method is missing endpoints in the current version of the web-API. These should be implemented.

# Requested changes

Implement the endpoints `/initialize` and `/navigate`, and associated [request/response/table models](https://desdeo.readthedocs.io/en/latest/explanation/database_models_and_schema/). There is an example found in `desdeo/api/routers/_NAUTILUS_navigator.py`, but this is an old implementation, and the new one should follow the conventions of the current web-API. The way E-NAUTILUS routers and models are implemented, should be followed when possible.

Instead of a sinlge PR, I would like to first see a PR with just the models. We can then move onto the routers. 

# Additional information

The [NAUTILUS Navigator paper](https://link.springer.com/article/10.1007/s10898-019-00765-2) should be read first to understand what the method is supposed to do.

PR #434, opened 2026-02-25
NAUTILUS Navigator endpoints - Models and schemas

PtrBednarik:
https://github.com/industrial-optimization-group/DESDEO/issues/422


PtrBednarik:
I made some changes, but I still do not consider it to be the final. 
I would like to know, if current version of 
desdeo/api/models/nautilus.py and desdeo/api/schemas/nautilus.py are correct.

Should I keep NautilusState in separate file (nautilus.py) or to move it among other states into models/state.py?

Second point is the file desdeo/api/models/generic_states.py. I have some questions here.

I imported new state there - NautilusState.
Added two new constants into class StateKind - 
    NAUTILUS_NAVIGATE = "nautilus.navigate"
    NAUTILUS_INITIALIZE = "nautilus.initialize"
Simillarly into KIND_TO_TABLE - 
    StateKind.NAUTILUS_NAVIGATE: NautilusState,
    StateKind.NAUTILUS_INITIALIZE: NautilusState

Problem comes with SUBSTATE_TO_KIND, where I wanted to do something like this :

    NautilusState: StateKind.NAUTILUS_NAVIGATE,
    NautilusState: StateKind.NAUTILUS_INITIALIZE
There is unfortunately collision of two values corresponding to the same NautilusState.
This is a problem.
I discussed a solution with ChatBot to modify create() method by adding a new argument kind: StateKind, what could solve it and remove the whole SUBSTATE_TO_KIND dict from the file.  It will also require explicit insertion of StateKind at every place the method StateDB.create() is used in the app. 
There is a commented sketch version of the "new" create() method as well.
I would like to discuss the solution, because I'm not sure with this intervention to the code logic.

PR #466, opened 2026-04-10
Added support for CVXPY

juropo:
Also fixed a problem with gurobipy not maximizing correctly.


gialmisi:
Check https://github.com/juropo/DESDEO/pull/15 @juropo 

I was checking #466 and noticed a couple of things worth addressing. Nothing breaking, but avoid a lot of duplication of code, for instance.

PR #470, opened 2026-04-20
Solution Comparison Table component added

MajoMartinco:
Solution Comparison Table component added. I am not exactly sure if Solution objects posses attribute 'name' but it should work anyway.


gialmisi:
Thanks, this is good for now!

PR #472, opened 2026-04-22
Stochastic optimization stuff

juropo:
This doesn't do that much yet. Only the representation for uncertainty has been implemented.

I will be working on adding support for mpi-sppy and maybe implementing some simple methods natively in DESDEO.

No one merge this yet.


gialmisi (commented):
I had a quick look at this and looks fine. I am happy to see the DB models have been given attention as well. The frontend will need some adjustments, since the problem definition page and some of the automatically generated API clients still refer to the old `scenario_keys`, but this is a small fix and can be addressed once we decide to merge this. I guess the ScenarioModelDB is supposed to be treated like a problem, so associating it with an interactive session makes probably no sense. Just checking.

I would like to see an example of how to utilize the scenarios at some point. I already got an initial idea from the updated river pollution problem, but having a documented use case example would be nice.

juropo:
Woo, tests are passing!

gialmisi (commented):
Why is `prices_summer.npz` bot at the root level and in the directory `datasets/`?

gialmisi (commented):
We should also start thinking of databse migration solutions. Dropping the `scenario_keys` would break all existing deployments, requiring the rebuild of the database instead. This is fine for now, but once we start having an 'official' deployed version, we need to addres this. This is a very general comment not related to this PR specifically.

juropo:
> Why is `prices_summer.npz` bot at the root level and in the directory `datasets/`?

By accident

desdeo/problem/json_parser.py
```diff
@@ -103,7 +106,76 @@ def __init__(self, to_format: FormatEnum = "polars"):  # noqa: C901
...
+            step = int(spec[2]) if len(spec) >= 3 else None  # noqa: PLR2004
+            start_0 = (start_1 - 1) if start_1 > 0 else (n + start_1)
+            stop_0 = (stop_1 - 1) if stop_1 > 0 else (n + stop_1)
+            if step is None:
+                step = 1 if stop_0 >= start_0 else -1
+            if step > 0:
+                return [i for i in range(start_0, stop_0 + 1, step) if 0 <= i < n]
+            return [i for i in range(start_0, stop_0 - 1, step) if 0 <= i < n]
+
+        def _collect_indices(raw_indices: tuple, n: int) -> list[int]:
```
gialmisi, line 109: What is `n` here?

desdeo/tools/robust.py
```diff
@@ -0,0 +1,307 @@
...
+                )
+            )
+
+        if found_type == "objectives":
+            new_objectives.append(
+                Objective(
+                    name=f"Min-max robust value of {sym}",
+                    symbol=robust_sym,
+                    func=t_sym,
+                    maximize=False,
```
gialmisi: Is this correct? For max problems, the value of the objective will be the negated _min value, which might be confusing if the value of this objective is ever meant to be read as is. Since the _min column is used anyways, having this match the original function's `maximize` should be no issue, but perhaps there is a reason for this I do not understand.

desdeo/tools/stochastic.py
```diff
@@ -0,0 +1,387 @@
...
+    return combined.model_copy(
+        update={
+            "objectives": new_objectives or None,
+            "scalarization_funcs": new_scal_funcs or None,
+            "extra_funcs": new_extra_funcs or None,
+        }
+    ), added_symbols
+
+
+def add_conditional_value_at_risk(
```
gialmisi, line 211: Calling this (and `add_min_max_robust`) on the same problem twice with the default prefixes will raise a Pydantic validation error. There could be an 'already exists' check to catch errors here as well, or document it at least that calling these twice with the same (or default) prefixes will raise an error.

desdeo/tools/stochastic.py
```diff
@@ -0,0 +1,387 @@
...
+        )
+
+    modified_model = scenario_model.with_base_problem(problem=scal_problem)
+    combined, symbol_maps = build_combined_scenario_problem(modified_model)
+    combined, added = add_expected_value(modified_model, [scal], combined=combined, symbol_maps=symbol_maps)
+
+    return combined, added[scal]
+
+
+def add_expected_value(
```
gialmisi, line 79: This is also problematic with calls on the same combiner problem. Aggregations in general would benefit from an 'already added'-check.

desdeo/tools/stochastic.py
```diff
@@ -0,0 +1,387 @@
...
+            problem=base_problem,
+            symbol=symbol,
+            reference_point=reference_point,
+            ideal=ideal,
+            nadir=nadir,
+            rho=rho,
+            delta=delta,
+        )
+
+    modified_model = scenario_model.with_base_problem(problem=scal_problem)
```
gialmisi: Since this skips model validation. One could also just construct a fresh ScenarioModel here to avoid possible corruptions.

desdeo/problem/scenario.py
```diff
@@ -0,0 +1,298 @@
...
+        return self.model_copy(update={"base_problem": new_base})
+
+    @cached_property
+    def leaf_scenarios(self) -> dict[str, float]:
+        """Return a dict mapping each leaf scenario name to its probability.
+
+        Leaf scenarios are nodes in the scenario tree with no children that also
+        appear in the scenarios dict.  If scenario_probabilities is empty, equal
+        weights (1/n) are assigned to each leaf.
+        """
```
gialmisi, line 264: Notice that the cached dict returned here is not a copy, but a reference, so any called can modify the dict, and hence the cache contents. If this is not intentional, consider returning a `MappingProxyType(dict)` instead.

desdeo/tools/scenarios.py
```diff
@@ -0,0 +1,488 @@
...
+def _extract_symbols(expr: str) -> set[str]:
+    """Return all variable/constant symbols referenced in a func string.
+
+    Uses the infix parser to produce a MathJSON AST, then walks it to collect
+    all string leaves that are not operator keywords.
+    """
+    parsed = _parser.parse(expr)
+    symbols: set[str] = set()
+    _walk_mathjson(parsed, symbols)
+    return symbols
```
gialmisi: Deadcode. Not used anywhere.

desdeo/tools/scenarios.py
```diff
@@ -0,0 +1,488 @@
...
+
+
+def _walk_mathjson(node, symbols: set[str]) -> None:
+    """Recursively collect non-operator string leaves from a MathJSON node."""
+    if isinstance(node, str):
+        if node not in _RESERVED:
+            symbols.add(node)
+    elif isinstance(node, list):
+        for child in node:
+            _walk_mathjson(child, symbols)
```
gialmisi: Deadcode together with `_extracts_symbols`

desdeo/problem/scenario.py
```diff
@@ -0,0 +1,298 @@
...
+            parent_prob = v.get(parent)
+            if parent_prob is None:
+                raise ValueError(f"Probability for node '{parent}' is missing.")
+            missing = [c for c in children if c not in v]
+            if missing:
+                raise ValueError(f"Probabilities missing for children of '{parent}': {missing}.")
+            child_sum = sum(v[c] for c in children)
+            if not math.isclose(child_sum, parent_prob, rel_tol=1e-9):
+                raise ValueError(f"Probabilities of children of '{parent}' sum to {child_sum}, expected {parent_prob}.")
+        return v
```
gialmisi, line 166: Only enforces "children sum to parent" when some child of a parent is in the dict. Should this include a full coverage instead?

gialmisi:
@juropo added some comments. Nothing breaking/critical I could find.

desdeo/problem/json_parser.py
juropo: It's the length of the array that we are looking at. It is necessary to support negative indices like `list[-1]` to reference the last value.

desdeo/tools/robust.py
juropo: Yeah, the only reason why maximize objectives was not being supported was the function name, but I changed it to worst case robust. It's still assuming that extra functions and scalarizations are minimized, because those have no direction indicated.

desdeo/tools/stochastic.py
juropo: I'm not sure how you even tried to do that. Pydantic validation error might be the correct response.

I added the correct way of accomplishing this in the documentation.

desdeo/tools/stochastic.py
juropo: I'm not sure they would benefit from that. I don't think we are checking if problem has scalarization functions added. I guess we could add a list called aggregations in the Problem class, but somehow that does not feel great either. I'd rather just leave this up to whoever is constructing methods.

desdeo/tools/stochastic.py
juropo: It's slower, but I guess it doesn't matter too much here. This is not a very good scalarization function anyways, and is mostly provided as an example.

I updated the with_base_problem function so that you can tell it to revalidate.

desdeo/problem/scenario.py
juropo: That probably shouldn't be allowed.

desdeo/tools/scenarios.py
juropo: It used to do something. Removed now.

desdeo/tools/scenarios.py
juropo: Removed

desdeo/problem/scenario.py
juropo: I instead added a validator that checks that the scenario tree is actually a tree. Then checking parent probabilities is enough. The two could have possibly been combined if it weren't for the fact that scenario probabilities are optional.

juropo:
Good review comments @gialmisi 
I updated the code accordingly.

PR #474, opened 2026-05-05
XLEMOO port to DESDEO

gialmisi:
Imported the XLEMOO method/framework to DESDEO using the current way of implementing EMO methods. I decided to create a new template for LEM-style MOO methods (learnable evolutionary methods).


desdeo/emo/methods/templates.py
```diff
@@ -109,3 +109,70 @@ def template2(
...
+            offspring_outputs = evaluator.evaluate(offspring)
+            solutions, outputs = selection.do(parents=(solutions, outputs), offsprings=(offspring, offspring_outputs))
+            if terminator.check():
+                stop = True
+                break
+        if stop:
+            break
+
+        for _ in range(n_learning_per_cycle):
+            solutions, outputs = learning_operator.do()
```
light-weaver: is this an entirely new set of solutions? no solutions kept from previous generations?

If there are any solutions left from the previous generations, I'd recommend not reevaluating them in the next line. Only evaluate unique solutions.

desdeo/emo/operators/selection.py
```diff
@@ -111,7 +111,7 @@ class ReferenceVectorOptions(BaseModel):
     """The method for normalizing the reference vectors. Defaults to "spherical"."""
     lattice_resolution: int | None = None
     """Number of divisions along an axis when creating the simplex lattice. This is not required/used for the "s_energy"
-    method. If not specified, the lattice resolution is calculated based on the `number_of_vectors`. If "spherical" is 
+    method. If not specified, the lattice resolution is calculated based on the `number_of_vectors`. If "spherical" is
     selected as the `vector_type`, this value overrides the `number_of_vectors`.
```
light-weaver: I don't remember why I wrote this but it seems wrong. Remove this line.

desdeo/emo/operators/selection.py
```diff
@@ -1776,3 +1774,141 @@ def state(self) -> Sequence[Message]:
...
+        """Topics this selector subscribes to (none)."""
+        return []
+
+    def __init__(
+        self,
+        problem: Problem,
+        verbosity: int,
+        publisher: Publisher,
+        population_size: int,
+        target_column: str,
```
light-weaver: I've been procrastinating on adding target_symbols as method arguments in other Selectors. Right now they symbols to minimize are chosen automagically in other selectors.

desdeo/emo/operators/selection.py
```diff
@@ -1776,3 +1774,141 @@ def state(self) -> Sequence[Message]:
 
     def update(self, message: Message) -> None:
         pass
+
+
+class ASFSelector(BaseSelector):
```
light-weaver: The implementation seems much more generic than "ASFSelector". It's basically an elitist single objective selector. Can you put this in `scalar_selection.py` intead? Despite the name the selectors in that file are used in MOEAs as well. The structure is _slightly_ different in that file. The `do` method expects a single tuple of (dec vars, outputs) which combine both parents and offsprings. It can optionally take a fitness array as well but you don't need to use that as all required data should already be in outputs for you.

desdeo/emo/operators/learning_mode.py
```diff
@@ -0,0 +1,156 @@
...
+    """Resolve an H/L split argument into a number of samples."""
+    if fraction < 1.0:
+        return int(fraction * n_total)
+    half = n_total // 2
+    if fraction > half:
+        return int(half)
+    return int(fraction)
+
+
+class LearningModeOperator:
```
light-weaver: If this is an operator, it should be a subscriber. You can implement new message topics for this class of operators

desdeo/emo/operators/learning_mode.py
```diff
@@ -0,0 +1,156 @@
...
+        self,
+        problem: Problem,
+        archive: Archive,
+        evaluator: EMOEvaluator,
+        selector: ASFSelector,
+        h_split: float = 0.2,
+        l_split: float = 0.2,
+        instantiation_factor: float = 10.0,
+        ml_model: SkopeRulesClassifier | None = None,
+        seed: int | None = None,
```
light-weaver: default seed is 0 for consistency with other operators

desdeo/emo/operators/learning_mode.py
```diff
@@ -0,0 +1,156 @@
...
+                n_estimators=30,
+                bootstrap=True,
+                bootstrap_features=True,
+                random_state=seed,
+            )
+        self.ml_model = ml_model
+        self.current_ml_model: SkopeRulesClassifier | None = None
+
+    def _latest_generation_population(self) -> tuple[pl.DataFrame, pl.DataFrame]:
+        """Return the latest archived generation as ``(decision_variables, outputs)``."""
```
light-weaver: Just a reminder. You are taking the latest offsprings, not the latest generation itself (which can include parents originally created an arbitrary number of generations before this step)

desdeo/emo/operators/learning_mode.py
```diff
@@ -0,0 +1,156 @@
...
+        Returns:
+            tuple[pl.DataFrame, pl.DataFrame]: Selected decision variables and their outputs.
+        """
+        target_column = self.selector.target_column
+        population_size = self.selector.population_size
+
+        all_solutions = self.archive.solutions
+        non_dec_cols = [c for c in all_solutions.columns if c not in (*self.variable_symbols, "generation")]
+
+        unique = all_solutions.unique(subset=self.variable_symbols, maintain_order=True)
```
light-weaver: huh. i wonder how expensive this operation becomes in later generations and problems with many dec vars

desdeo/emo/operators/learning_mode.py
```diff
@@ -0,0 +1,156 @@
...
+    ):
+        """Initialize the learning-mode operator.
+
+        Args:
+            problem (Problem): The optimization problem.
+            archive (Archive): Archive holding all evaluated solutions so far.
+            evaluator (EMOEvaluator): Evaluator used to score newly instantiated individuals.
+            selector (ASFSelector): ASF selector that defines the target column and the
+                desired population size.
+            h_split (float, optional): Fraction (or absolute count if ``>= 1``) of unique
```
light-weaver: this makes things so much more confusing than they need to be. I had a similar issue in some other EMOs. In the end, I just just one of the options (frac or count) arbitrarily and stuck to it.

desdeo/emo/operators/learning_mode.py
```diff
@@ -0,0 +1,156 @@
...
+        non_dec_cols = [c for c in all_solutions.columns if c not in (*self.variable_symbols, "generation")]
+
+        unique = all_solutions.unique(subset=self.variable_symbols, maintain_order=True)
+        sorted_unique = unique.sort(target_column)
+
+        n_total = sorted_unique.height
+        h_size = _split_indices(n_total, self.h_split)
+        l_size = _split_indices(n_total, self.l_split)
+
+        if h_size == 0 or l_size == 0:
```
light-weaver: I assume this is because h_split or l_split is 0. Otherwise if n_total is 0, then latest gen should also be empty by definition and everything will break. If the assumption is right, you can probably put this at the top of the method and skip unnecessary calculations

desdeo/emo/operators/learning_mode.py
```diff
@@ -0,0 +1,156 @@
...
+        l_vars = l_group[self.variable_symbols].to_numpy()
+
+        x_train = np.vstack((h_vars, l_vars))
+        y_train = np.hstack((np.ones(len(h_vars), dtype=int), np.zeros(len(l_vars), dtype=int)))
+
+        self.current_ml_model = self.ml_model.fit(x_train, y_train, feature_names=self.variable_symbols)
+
+        rules, weights = extract_skoped_rules(self.current_ml_model)
+        if not rules:
+            return self._latest_generation_population()
```
light-weaver: This comment applies to other similar returns. Are you _sure_ you want the algorithm to continue with the latest offsprings instead of the latest generation? The latest offsprings can be arbitrarily bad.

I will send a pull request about the other "observer" hook soon, that might be a better hook for this task instead of using an archiver

desdeo/emo/operators/learning_mode.py
```diff
@@ -0,0 +1,156 @@
...
+            rules_list=rules,
+            weights=weights,
+            variable_symbols=self.variable_symbols,
+            variable_bounds=self.variable_bounds,
+            n_samples=n_to_instantiate,
+            rng=self.rng,
+        )
+
+        instantiated_df = pl.DataFrame(instantiated, schema=self.variable_symbols)
+        instantiated_outputs = self.evaluator.evaluate(instantiated_df)
```
light-weaver: You have the evaluation step twice. Once here and once in the template. The archiver (or any other objected hooked into the evaluator) will see these solutions twice. There is technically a selection step in the middle but the second set is still a strict subset of the first set.

light-weaver: I recommend returning `instantiated_df` at line 148. Then do evaluation (149) and selection (153) in the template. It will also make things easier to understand. The learning mode operator finds a good set of decision vars for the algorithm, then the algorithm can do whatever it wants with that information.

desdeo/emo/operators/learning_mode.py
light-weaver: though, technically, you don't even need to use the new hook.

If you want to use the old generation as is, just return None or some error. Catch that error/None in the template, and use whatever `solutions`/`outputs` already exist in the template. They should contain the old generation solutions.

desdeo/emo/operators/learning_mode.py
light-weaver: Well... come to think of it, if h_split or l_split is 0 then this operator never does anything, so this case should never occur in the first place I guess.

desdeo/emo/methods/templates.py
```diff
@@ -109,3 +109,70 @@ def template2(
...
+        n_learning_per_cycle (int, optional): Number of Learning iterations per cycle. Set
+            to 0 to disable Learning mode entirely. Defaults to 1.
+
+    Returns:
+        EMOResult: The final population and its objective/target values.
+    """
+    solutions, outputs = generator.do()
+
+    while not terminator.check():
+        stop = False
```
light-weaver: i think you just reinvented goto

desdeo/emo/operators/learning_mode.py
light-weaver: This would also decouple this operator from the "ASFSelector". You should technically be able to use any MOEA selector with learningmode operator. The selectors could keep doing multiobjective selection using objective values and this operator could use the ASF value in the outputs df to do its task.

desdeo/emo/options/templates.py
```diff
@@ -373,11 +446,80 @@ def emo_constructor(
...
     template_funcs = {
         "Template1": template1,
         "Template2": template2,
+        "TemplateXLEMOO": template_xlemoo,
     }
 
+    if template.name == "TemplateXLEMOO":
+        if not isinstance(selector, ASFSelector):
+            raise InvalidTemplateError("XLEMOO requires an ASFSelector.")
+        learning_operator = LearningModeOperator(
```
light-weaver: this should be a subscriber. consistency checks should happen after everything is connected to each other

light-weaver:
You don't need to have the archive as one of the init arguments. Instead, `LearningModeOperator` could subscribe to the same messages as the archive, and maintain its own (but much smaller and more efficient) archive -> set of `h_group` and `l_group` solutions. By doing this you would only need to do the `unique()` operation on a single generation's worth of solutions (plus the h and l groups) at each `update()` call, instead of running the same operation on a potentially unbounded archive on each `do()` call.

This also makes the construction of the template simpler. Now you no longer need to assume that an archive is connected. Instead, the user can choose to connect whatever kind of archive they want.

gialmisi:
@light-weaver could you give a quick check? I think to have addressed most issues now.

desdeo/emo/options/algorithms.py
light-weaver, line 482: you can add xlmoo options here if you want to dump the schema to the disk. might be useful in the future for frontend purposes idk?

desdeo/emo/options/templates.py
```diff
@@ -250,7 +313,7 @@ def preference_handler(
             problem=problem,
             aspiration_levels=preference.aspiration_levels,
             reservation_levels=preference.reservation_levels,
-            desirability_levels={name: preference.desirability_levels for name in preference.aspiration_levels},
+            desirability_levels=dict.fromkeys(preference.aspiration_levels, preference.desirability_levels),
```
light-weaver, line 316: ryanrenoldsbutwhy.gif

desdeo/explanations/rules.py
light-weaver, line 1: skipped this file

light-weaver (commented):
Looks good. Merge whenever you resolve the conflicts

desdeo/emo/options/templates.py
gialmisi: Because ruff.

PR #479, opened 2026-05-07
Rahti deployment adjustments

viljokass:
at last getting back to this: adjust build configs to point to master and small fix to webui dockerfile


juropo:
I think the current version of master branch has an API that is incompatible with the webui. See https://github.com/industrial-optimization-group/DESDEO/pull/480

gialmisi:
Can be merged.

PR #480, opened 2026-05-08
Removed scenario keys from webui

juropo:
This should make the webui compatible with the changes that were made to the api when scenario support was added.


juropo:
I don't think there's anything controversial here.

Issue #492, opened 2026-06-03
Audit and fix broken/incorrect links in the documentation


gialmisi:
# Background

Issue #281 notes that the documentation has some wrong links. Broken
links are not very nice to have in the documentation.

# What to do

- Pick one documentation area to keep the task small (e.g. the `docs/tutorials/`
  folder, or the `docs/howtoguides/` folder). **Each contributor can pick one area to work on.**
- Read through the Markdown/notebook files and check every internal link
  (`../something.md#anchor`) and external link.
- Fix the ones that 404 or point to the wrong page/anchor. Cross-reference #281 in
  your PR description.
- Optionally, build the docs locally (`mkdocs serve`) and click through your section
  to confirm the links resolve.

# Done when

- All links in the chosen section resolve correctly.
- The PR references issue #281 and lists which links were fixed.

# Resources

- The documentation has pointers on how to build and check the documentation locally in the 'Contributing' section.

Issue #281, opened 2025-06-12
Documentation has some wrong links


gialmisi:
The documentation should be checked for its hyperlinks. For instance, the first hyperlink here: https://desdeo.readthedocs.io/en/latest/howtoguides/nautilus_navigator/

Points to a markdown file, whereas it should point to https://desdeo.readthedocs.io/en/latest/howtoguides/how_to_define_a_problem/

The links on the documentation pages should be systematically checked, and incorrect links should be fixed. If it is not clear what a link should point to, but it is clearly incorrect, you can ask below for clarifications.

PR #507, opened 2026-06-05
Fixed links in the "Explanation" section of the documentation

Mastereeli:
Fixed links in the "Explanation" section of the documentation.
- Some links had a harmful prefix in the fragments and required dashes instead of underscores (#**desdeo.tools.**scipy_solver_interfaces -> #scipy-solver-interfaces)
- One link was missing a slash (/)
- "Supported expressions" -> "Supported operations"
- Removed a file extension from the end of a file name in a link

Swapped around Models and schema and Group decision making and Websockets in the navigation menu to reflect the order in which they appear in the Explanation index.

It should be noted that Sympy Evaluator and Gurobipy Evaluator don't appear in latest/api/desdeo_problem, so links to those cannot work.

#281
#492


Mastereeli:
The pre-commit complains on naming conventions for variables in example code inside the documentation.

PR #511, opened 2026-06-08
Fixing mypy

Ukez1:


desdeo/tools/interaction_schema.py
```diff
@@ -3,6 +3,7 @@
 from pydantic import BaseModel, Field
 
 
+# ruff: noqa: T201
```
gialmisi: If you check the current master, you will see that the error skip has been moved to the actual print statements. Can you move this below next to the print statements as well?

gialmisi: Actually, just update this whole file to match the current master branch one.

gialmisi (commented):
Try to pull the current master branch and merge it into this, should fix most issues right away.

gialmisi (changes requested):
@Ukez1

gialmisi (approved):
Good!

Issue #502, opened 2026-06-04
Add the ZDT4 test problem


gialmisi:
# Background
DESDEO has ZDT1/2/3 but not ZDT4. ZDT4 is a standard multimodal benchmark from the same paper. Template to copy: `zdt1` in `zdt_problem.py`.
 
# What to do
- Add a `zdt4(number_of_variables: int) -> Problem` function in `zdt_problem.py`.
- Implement from the reference. ZDT4 (minimize both): `f_1 = x_1`; `g = 1 + 10(n−1) + sum_{i=2}^n [x_i**2 − 10*Cos(4*pi*x_i)]`; `f_2 = g*(1 − Sqrt(f_1/g))`. Variable bounds: `x_1 in [0,1]`, `x_2…x_n in [−5,5]`.
- Register it in `__init__.py` and add a test asserting objective values at a known point (e.g. all variables at their lower bound, or `x_1=0.5`, rest `=0`).
- **Verify your formulas against the paper before trusting your test numbers.**

# Reference
Zitzler, E., Deb, K., & Thiele, L. (2000). *Comparison of multiobjective evolutionary algorithms: Empirical results.* Evolutionary Computation, 8(2), 173–195.
 
# Done when
- `zdt4` is importable from `desdeo.problem.testproblems`, evaluates correctly, and has a passing test.

PR #515, opened 2026-06-10
Added the ZDT4 problem and tests for it.

Mastereeli:
#502

### DIFF
--- a/desdeo/problem/testproblems/zdt_problem.py
+++ b/desdeo/problem/testproblems/zdt_problem.py
@@ -1,3 +1,5 @@
+"""Contains implementations of ZDT problems."""
+
 from math import pi
 
 from desdeo.problem.schema import (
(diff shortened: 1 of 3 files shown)


desdeo/problem/testproblems/zdt_problem.py
```diff
@@ -272,3 +274,93 @@ def zdt3(
...
+        \min\quad f_1(\textbf{x}) &= x_1 \\
+        \min\quad f_2(\text{x})\&=g(\text{x})\cdot\left(h\right) \\
+        g(\text{x})\&=1+10\left(n-1\right)+\sum_{i=2}^n\left[x_i^2-10\cos\left(4\pi\cdot x_i\right)\right] \\
+        h(f_1{,}g)\&=1-\sqrt{\frac{f_1}{g}}, \\
+    \end{align*}
+
+    where $f_1$ and $f_2$ are objective functions, $x_1,\dots,x_n$ are decision variables, $n$
+    is the number of decision variables,
+    and $g$ and $h$ are auxiliary functions.
+    """
```
gialmisi, line 295: The rendered LaTeX if funky. Check it by running `just docs-fast` and then navigating to http://127.0.0.1:8000/en/latest/api/desdeo_problem/#desdeo.problem.testproblems.zdt4 compare the rendering to, e.g., that of ZDT3.

tests/test_testproblems.py
```diff
@@ -469,6 +470,28 @@ def test_mcwb_ragsdell1976_problem():
     assert np.isclose(f2, 1.2e-06, rtol=1e-3, atol=1e-9)
 
 
+@pytest.mark.eeli
```
gialmisi: Perhaps you meant `testproblem`?

gialmisi (changes requested):
Check the rendering of the LaTeX expression for ZDT4.

tests/test_testproblems.py
Mastereeli: Forgot to remove that. I used it to specifically run only that test :D

gialmisi (approved):
Good!

PR #516, opened 2026-06-10
Remove broken and malicious script

lukarantalainen:
Removed a script that was causing the annoying fake login prompt to trigger on the desdeo docs website.

[https://www.kaspersky.com/blog/polyfill-io-service-supply-chain-attacks/51635/](https://www.kaspersky.com/blog/polyfill-io-service-supply-chain-attacks/51635/)

image

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -157,5 +157,4 @@ extra_css:
 extra_javascript:
   - extrajs/extra.js
   - extrajs/mathjax.js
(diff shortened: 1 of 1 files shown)


gialmisi (approved):
Very good catch! This was already a redundant addition as it was supposed to support very old browsers. It originated from MathJax's 'getting started' snippet, which is a good reminder to always check the reasoning between every line even in "official" snippets.

Issue #505, opened 2026-06-04
Add the DTLZ4 test problem


gialmisi:
# Background
DTLZ4 is DTLZ2 with the position variables raised to a power `alpha` (commonly `alpha = 100`), which biases the spread of solutions.
 
# What to do
- Copy `dtlz2` to a new `dtlz4(n_variables, n_objectives, alpha: float = 100.0)` function.
- Replace each position variable `x_i` in the objective expressions with `x_i**alpha`, following the reference.
- Register in `__init__.py`; add a test asserting values at a known point.

# Reference

Deb, K., Thiele, L., Laumanns, M., & Zitzler, E. (2005). *Scalable test problems for evolutionary multiobjective optimization.* In Evolutionary Multiobjective Optimization (pp. 105–145). Springer.

# Done when
- `dtlz4` is importable, evaluates correctly, and has a passing test.

PR #517, opened 2026-06-10
Added DTLZ4 implementation and fixed some other issues

lukarantalainen:
Added DTLZ4 problem and tests for it. Fixed a bug in DTLZ2 test where the test result depended on the previous test. Also changed the code in DTLZ2 g function to better match the definition in docstring (no change in function).
#505


desdeo/problem/testproblems/dtlz_problems.py
```diff
@@ -49,17 +52,16 @@ def dtlz2(n_variables: int, n_objectives: int) -> Problem:
...
     for m in range(1, n_objectives + 1):
         # function f_m
         prod_expr = " * ".join([f"Cos(0.5 * {np.pi} * x_{i})" for i in range(1, n_objectives - m + 1)])
         if m > 1:
-            prod_expr += f"{' * ' if prod_expr != "" else ""}Sin(0.5 * {np.pi} * x_{n_objectives - m + 1})"
+            prod_expr += f"{' * ' if prod_expr != '' else ''}Sin(0.5 * {np.pi} * x_{n_objectives - m + 1})"
         if prod_expr == "":
             prod_expr = "1"  # When m == n_objectives, the product is empty, implying f_M = g.
-        f_m_expr = f"({g_symbol}) * ({prod_expr})"
+        f_m_expr = f"(1 + {g_symbol}) * ({prod_expr})"
```
gialmisi, line 64: This is fine!

desdeo/problem/testproblems/dtlz2_problem.py
```diff
@@ -100,3 +102,96 @@ def dtlz2(n_variables: int, n_objectives: int) -> Problem:
...
+    g_expr = " + ".join([f"(x_{i} - 0.5)**2" for i in range(n_objectives, n_variables + 1)])
+
+    objectives = []
+    for m in range(1, n_objectives + 1):
+        # function f_m
+        prod_expr = " * ".join([f"Cos(0.5 * {np.pi} * x_{i}**{alpha})" for i in range(1, n_objectives - m + 1)])
+        if m > 1:
+            prod_expr += f"{' * ' if prod_expr != '' else ''}Sin(0.5 * {np.pi} * x_{n_objectives - m + 1}**{alpha})"
+        if prod_expr == "":
+            prod_expr = "1"  # When m == n_objectives, the product is empty, implying f_M = g.
```
gialmisi:
```suggestion
            prod_expr = "1"  # Only reached when n_objectives == 1: no trig terms, so f_1 = 1 + g.
```

gialmisi (changes requested):
Good, the problem formulation seems correct. One change: can you rename the from `dtlz2_problem.py` to `dtlz_problems.py` since we now have more than just DTLZ2 defined? This would warrant a change in related imports as well.

PR #518, opened 2026-06-11
Implemented dtlz1 test problem and wrote tests for it

lukarantalainen:


desdeo/problem/testproblems/dtlz_problems.py
```diff
@@ -11,6 +11,109 @@
 )
 
 
+def dtlz1(n_variables: int, n_objectives: int) -> Problem:
```
gialmisi, line 14: Can you add a M=1 guard?

desdeo/problem/testproblems/dtlz_problems.py
```diff
@@ -11,6 +11,109 @@
...
+    \end{equation}
+
+    \begin{equation}
+        f_m(\mathbf{x}) = \frac{1}{2}(1+g(\mathbf{x}_M))
+        \prod_{j=1}^{M-m} x_j (1 - x_{M-m+1}), \quad m=2,\dots,M-1
+    \end{equation}
+
+    \begin{equation}
+        f_M(\mathbf{x}) = \frac{1}{2}(1+g(\mathbf{x}_M))(1 - x_1)
+    \end{equation}
```
gialmisi: The rendering of the equations could be nicer, plus the 'min'-operator is not present. See, for instance, how the R22 problem's (`desdeo/problem/testproblems/re_problem.py`) LaTeX is typed for some guidance, e.g., use the `align` environment instead of multiple `equation` ones.

gialmisi (changes requested):
The problem seems to be defined correctly, some minor issues still to fix. See my comments.

gialmisi (approved):

Issue #519, opened 2026-06-11
Fix griffe warnings when building docs


gialmisi:
# Background

When building the docs of DESDEO (e.g., `just docs-fast`) `griffe` outputs a bunch of warnings. While these are not breaking issues, they are still issue that affect the quality of the code and the documentation, and should there fore be fixed.

# What to do

Fix at least the sources for the following warnings:

```
WARNING -  griffe: desdeo/problem/scenario.py:245: No type or annotation for parameter '**updates'
WARNING -  griffe: desdeo/problem/gurobipy_evaluator.py:275: Parameter 'model' does not appear in the function signature
WARNING -  griffe: desdeo/problem/gurobipy_evaluator.py:584: Failed to get 'name: description' pair from 'function already found in the model.'
WARNING -  griffe: desdeo/problem/cvxpy_evaluator.py:442: No type or annotation for parameter '**kwargs'
WARNING -  griffe: desdeo/tools/group_scalarization.py:55: No type or annotation for parameter 'agg_bounds dict[str, float]'
WARNING -  griffe: desdeo/tools/group_scalarization.py:55: Parameter 'agg_bounds dict[str, float]' does not appear in the function signature
WARNING -  griffe: desdeo/tools/group_scalarization.py:181: Parameter 'reference_points' does not appear in the function signature
WARNING -  griffe: desdeo/tools/group_scalarization.py:457: Parameter 'reference_points' does not appear in the function signature
WARNING -  griffe: desdeo/tools/group_scalarization.py:458: No type or annotation for parameter 'agg_bounds dict[str, float]'
WARNING -  griffe: desdeo/tools/group_scalarization.py:458: Parameter 'agg_bounds dict[str, float]' does not appear in the function signature
WARNING -  griffe: desdeo/tools/group_scalarization.py:302: No type or annotation for parameter 'agg_bounds dict[str, float]'
WARNING -  griffe: desdeo/tools/group_scalarization.py:302: Parameter 'agg_bounds dict[str, float]' does not appear in the function signature
WARNING -  griffe: desdeo/tools/group_scalarization.py:2403: No type or annotation for parameter 'agg_bounds dict[str, float]'
WARNING -  griffe: desdeo/tools/group_scalarization.py:2403: Parameter 'agg_bounds dict[str, float]' does not appear in the function signature
WARNING -  griffe: desdeo/tools/group_scalarization.py:2531: Parameter 'reference_points' does not appear in the function signature
[...]

PR #520, opened 2026-06-11
No more warnings when building docs and fixed some ruff complaints.

Mastereeli:
Fixed the griffe warnings #519 
Tried to please Ruff, but it complains too much. I'm not paid enough to rewrite Nimbus functions to be less complex :DD
Also, after going through what I was able to fix for Ruff, it came up with more complaints.
Renamed GenerateReferencePoints.py to generateReferencePoints.py
About 8 tests now pass that were caused by what seemed to be incomplete changing of how the gurobipy_evaluator.init_constraints worked. The documentation and some of the calls for it seem to be outdated though.

### DIFF
--- a/desdeo/problem/cvxpy_evaluator.py
+++ b/desdeo/problem/cvxpy_evaluator.py
@@ -435,11 +435,11 @@ class CVXPYEvaluator:
         self.objective_expr = objective
         self.problem_model = cp.Problem(objective, list(self.constraints.values()))
 
-    def solve(self, **kwargs):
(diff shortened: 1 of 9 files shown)


Mastereeli:
Is this fine? I'm pretty sure I did not introduce the problems the pre-commit and Unit tests are flagging.
Especially the flagged unit tests seem to be exactly the same ones that failed before my changes. Pre-commit flags are probably caused by the amount of files I made small changes to.

gialmisi (approved):
Good!

Issue #528, opened 2026-06-12
Add the water management (water resource planning) problem


gialmisi:
# Background
The water resource planning problem is a 5-objective, 3-variable constrained engineering benchmark: it trades off drainage network cost and several pollution/quality measures, subject to seven constraints. It is a popular multiobjective test problem with analytic expressions. (Note: this is distinct from the already-implemented River Pollution problem.)

Template to copy: `desdeo/problem/testproblems/re_problem.py`.

# What to do
- [ ] Add `water_management_problem() -> Problem` in `desdeo/problem/testproblems/water_management_problem.py` with the 3 decision variables, 5 objectives, and 7 constraints from the reference.
- [ ] Write the objective/constraint expressions in the DESDEO form, with the formulas in the docstring (LaTeX).
- [ ] Register in `__init__.py`; add a test asserting values at a known feasible point.

# Reference
Ray, T., Tai, K., & Seow, K. C. (2001). *Multiobjective design optimization by an evolutionary algorithm.* Engineering Optimization, 33(4), 399–424. (Section 4.4.)

# Done when
- [ ] `water_management_problem` is importable, evaluates correctly (objectives and constraints), and has a passing test.

PR #531, opened 2026-06-15
Making water management problem

Ukez1:
Created water management problem based on the Add the water management (water resource planning) problem
 #528

### DIFF
--- a/desdeo/problem/testproblems/water_management_problem.py
+++ b/desdeo/problem/testproblems/water_management_problem.py
@@ -51,7 +51,7 @@ def water_management() -> Problem:
         variable_type=VariableTypeEnum.real,
         lowerbound=0.01,
         upperbound=0.10,
-        initial_value=0.2,
+        initial_value=0.05,
     )
 
     x_3 = Variable(
@@ -60,7 +60,7 @@ def water_management() -> Problem:
         variable_type=VariableTypeEnum.real,
         lowerbound=0.01,
         upperbound=0.10,
-        initial_value=0.2,
+        initial_value=0.05,
     )
 
     # Constraints
@@ -146,7 +146,7 @@ def water_management() -> Problem:
     f_3 = Objective(
         name="f_3",
         symbol="f_3",
-        func="((305700) * 2289 * x_2)/((0.06*2289)**0.65)",
+        func="(305700 * 228.9 * x_2)/((0.06*2289)**0.65)",
         objective_type=ObjectiveTypeEnum.analytical,
         is_linear=False,
         is_convex=False,
@@ -156,7 +156,7 @@ def water_management() -> Problem:
     f_4 = Objective(
         name="f_4",
         symbol="f_4",
-        func="250 * 2289 * np.exp(-39.75 * x_2 + 9.9 * x_3 + 2.74)",
+        func="250 * 2289 * Exp(-39.75 * x_2 + 9.9 * x_3 + 2.74)",
         objective_type=ObjectiveTypeEnum.analytical,
         is_linear=False,
         is_convex=False,
--- a/desdeo/problem/testproblems/water_management_problem.py
+++ b/desdeo/problem/testproblems/water_management_problem.py
@@ -17,6 +17,16 @@ def water_management() -> Problem:
     The constraints and objective functions for the water management problem are defined as follows:
 
     \begin{align}
+        \min\quad f_1(\textbf{x})
+        &= 106780.37(x_2 + x_3) + 61704.67 \\
+        \min\quad f_2(\textbf{x})
+        &= 3000x_1 \\
+        \min\quad f_3(\textbf{x})
+        &= \frac{305700 \cdot 2289 x_2} {(0.06 \cdot 2289)^{0.65}} \\
+        \min\quad f_4(\textbf{x})
+        &= 250 \cdot 2289 \cdot \exp(-39.75x_2 + 9.9x_3 + 2.74) \\
+        \min\quad f_5(\textbf{x})
+        &= 25\left( \frac{1.39}{x_1x_2} + 4940x_3 - 80 \right)
         g_1(\mathbf{x})
         &= \frac{0.00139}{x_1 x_2} + 4.94x_3 - 0.08 \leq 1.00 \\
         g_2(\mathbf{x})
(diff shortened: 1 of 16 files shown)


gialmisi (changes requested):
@Ukez1 could you also add a test that would check the problem at some known points? The article provides the following handy table:

Image

Having a test that checks that the problem evaluates to the same objective function values with the given decision variables values in the table would be a very solid addition.

Source of image: Ray, T., Tai, K., & Seow, K. C. (2001). Multiobjective design optimization by an evolutionary algorithm. Engineering Optimization, 33(4), 399–424. (Section 4.4., Table III)

desdeo/problem/testproblems/water_management_problem.py
```diff
@@ -0,0 +1,190 @@
...
+        initial_value=0.2,
+    )
+
+    x_2 = Variable(
+        name="x_2",
+        symbol="x_2",
+        variable_type=VariableTypeEnum.real,
+        lowerbound=0.01,
+        upperbound=0.10,
+        initial_value=0.2,
```
gialmisi: Check that the initial values of the variables are withing the bounds.

desdeo/problem/testproblems/water_management_problem.py
```diff
@@ -0,0 +1,190 @@
...
+    )
+
+    # Constraints
+
+    g_1 = Constraint(
+        name="g_1",
+        symbol="g_1",
+        cons_type=ConstraintTypeEnum.LTE,
+        func="(0.00139/(x_1 * x_2))+ 4.94 * x_3 -0.08 - 1.00",
```
gialmisi: Having the minus sign `-0.08` instead of `- 0.08` (notice the space)  can be misinterpreted as a negation instead of a minus operation. I suggest having a space between the sign and the number when the minus operation is in question. Check this is all expressions.

desdeo/problem/testproblems/water_management_problem.py
```diff
@@ -0,0 +1,190 @@
...
+        objective_type=ObjectiveTypeEnum.analytical,
+        is_linear=False,
+        is_convex=False,
+        is_twice_differentiable=True,
+    )
+
+    f_3 = Objective(
+        name="f_3",
+        symbol="f_3",
+        func="((305700) * 2289 * x_2)/((0.06*2289)**0.65)",
```
gialmisi: The first term in the `func` expression has extra parentheses, they can be removed.

desdeo/problem/testproblems/water_management_problem.py
```diff
@@ -0,0 +1,190 @@
...
+        objective_type=ObjectiveTypeEnum.analytical,
+        is_linear=False,
+        is_convex=False,
+        is_twice_differentiable=True,
+    )
+
+    f_4 = Objective(
+        name="f_4",
+        symbol="f_4",
+        func="250 * 2289 * np.exp(-39.75 * x_2 + 9.9 * x_3 + 2.74)",
```
gialmisi: The `np.exp` will not work as it is not a recognized operation in DESDEO's infix syntax. Check the supported operations in [the documentation](https://desdeo.readthedocs.io/en/latest/explanation/problem_format/#supported-operations). You will want to use `Exp()` here.

tests/test_testproblems.py
```diff
@@ -679,3 +680,25 @@ def test_zdt6():
...
     assert np.isclose(g, 8.568067737283432)
+
+
+@pytest.mark.testproblem
+def test_water_management():
+    """Test that water management problem evaluates correctly."""
+    problem = water_management()
+
+    evaluator = PolarsEvaluator(problem)
+    xs = {"x_1": [0.1312], "x_2": [0.0942], "x_3": [0.0354]}
```
gialmisi: The test should check all four cases presented in the table I added in my previous comment to the PR.

gialmisi (changes requested):
Some issues to still take care of. See my comments.

desdeo/problem/testproblems/water_management_problem.py
```diff
@@ -0,0 +1,190 @@
...
+    Problem,
+    Variable,
+    VariableTypeEnum,
+)
+
+
+def water_management() -> Problem:
+    r"""Water resource planning problem.
+
+    The objective functions and constraints for the water management problem are defined as follows:
```
gialmisi: I noticed that the constraints are absent from the LaTeX:

Image

desdeo/problem/testproblems/water_management_problem.py
```diff
@@ -0,0 +1,190 @@
...
+
+    f_1 = Objective(
+        name="f_1",
+        symbol="f_1",
+        func="106780.37 * (x_2 + x_3) + 61704.67",
+        objective_type=ObjectiveTypeEnum.analytical,
+        is_linear=False,
+        is_convex=False,
+        is_twice_differentiable=True,
+    )
```
gialmisi: All functions in this problem (both objectives and constraints) **are** convex. Some are linear, some are not, I can check this in the final PR (but you are free to give it a go as well, if you wish). When it comes to differentiability, the functions are marked correctly.

gialmisi (changes requested):
Good, the problem is now defined correctly, but there are some definitions missing in the docstring and the properties of the functions need to be checked (partly by me on the final PR). Check my comments.

desdeo/problem/testproblems/water_management_problem.py
Ukez1: I wrote the constraints in LaTex, they should be right and accurate. I also checked the functions properties. I fixed the ones which I was completely sure, but someones I didn't edit.

desdeo/problem/testproblems/water_management_problem.py
```diff
@@ -0,0 +1,204 @@
...
+        g_3(\mathbf{x})
+        &= \frac{12.307}{x_1 x_2} + 49408.24x_3 + 4051.02 \leq 50000.00 \\
+        g_4(\mathbf{x})
+        &= \frac{2.098}{x_1 x_2} + 8046.33x_3 - 696.71 \leq 16000.00 \\
+        g_5(\mathbf{x})
+        &= \frac{2.138}{x_1 x_2} + 7883.39x_3 - 705.04 \leq 10000.00 \\
+        g_6(\mathbf{x})
+        &= \frac{0.417}{x_1 x_2} + 1721.26x_3 - 136.54 \leq 2000.00 \\
+        g_7(\mathbf{x})
+        &= \frac{0.164}{x_1 x_2} + 631.13x_3 - 54.48 \leq 550.00 \\
```
gialmisi, line 43: List the constraints after the objective functions. Indicate the range of the variables as well.

gialmisi (changes requested):
Check the LaTeX. Try rendering it by running `just docs-fast` and navigating to `http://127.0.0.1:8000/en/latest/api/desdeo_problem/#desdeo.problem.testproblems.water_management` (or wherever the local server is hosted). It currently renders in a funky way. Check some of the existing test problems with correct LaTeX rendering for tips on how to adjust the LaTeX source.

Issue #530, opened 2026-06-15
Unused ideal in group guess scalarization function


gialmisi:
@juropo I think the ideal point is unneeded here. Are these lines safe to delte? Please comment.

https://github.com/industrial-optimization-group/DESDEO/blob/a1cf816fc224e2a2fb3a24af3666dbfda3a56dfb/desdeo/tools/group_scalarization.py#L2548-L2556

juropo:
Yes, they should be safe to delete. @jpajasmaa might actually be more knowledgeable about these, even though git blame is pointing at me.

jpajasmaa:
Yes, I do not think it is used there and it some left over from some earlier version I guess. It can be deleted.

gialmisi:
@Mastereeli @lukarantalainen @Ukez1 Can one of you remove the lines related to the ideal point I reference in the OP? Thanks!

gialmisi:
Addressed in #537 closing

PR #537, opened 2026-06-18
removed unnecessary lines

lukarantalainen:
#530


lukarantalainen:
Perhaps the ideal parameter should be removed from the function too? It would probably cause some issues though...

gialmisi (approved):

PR #538, opened 2026-06-18
Missing docstrings

Ukez1:
Fixed some missing docstrings that ruff reported. Also split too long lines and fixed typos in comments that I noticed.


gialmisi (changes requested):
I believe this has the same changes as #531 ? If so, this should be closed.

gialmisi (changes requested):
Check the comments I gave in https://github.com/industrial-optimization-group/DESDEO/pull/531

PR #546, opened 2026-06-23
Adding the car side impact problem

Ukez1:


desdeo/problem/testproblems/car_side_impact_problem.py
```diff
@@ -0,0 +1,370 @@
...
+        &+ \max(0.261 - 0.0159x_1x_2 - 0.06486x_1 - 0.019x_2x_7
+        + 0.0144x_3x_5 + 0.0154464x_6 - 0.32, 0) \\
+        &+ \max(0.74 - 0.61x_2 - 0.031296x_3 - 0.031872x_7 + 0.227x_2^2 - 0.32,0) \\
+        &+ \max(0.74 - 0.61x_2 - 0.031296x_3 - 0.031872x_7 + 0.227x_2^2 - 0.32,0) \\
+        &+ \max(28.98 + 3.818x_3 - 4.2x_1x_2 + 1.27296x_6 - 2.68065x_7 - 32.0,0) \\
+        &+ \max(33.86 + 2.95x_3 - 5.057x_1x_2 - 3.795x_2 - 3.4431x_7 + 1.45728 - 32.0,0) \\
+        &+ \max(46.36 - 9.9x_2 - 4.4505x_1 - 32.0,0) \\
+        &+ \max(4.72 - 0.5x_4 - 0.19x_2x_3 - 4.0, 0) \\
+        &+ \max(10.58 - 0.674x_1x_2 - 0.67275x_2 - 9.9, 0) \\
+        &+ \max(16.45 - 0.489x_3x_7 - 0.843x_5x_6 - 15.7, 0)
```
gialmisi: Line 31 is repeated twice, thus f4 does not match the code. (the code is correct)

Line 38 is missing a `\\` at the end.

tests/test_testproblems.py
```diff
@@ -717,3 +718,49 @@ def test_water_management():
...
+
+    xs = {
+        "x_1": 0.5,
+        "x_2": 0.45,
+        "x_3": 0.5,
+        "x_4": 0.5,
+        "x_5": 0.875,
+        "x_6": 0.4,
+        "x_7": 0.4,
+    }
```
gialmisi, line 739: You test for the lower bounds, good! Could you also test for the upper bounds and middle values (`(upper bound - lower bound) / 2`)?

gialmisi (changes requested):
Good! The definition of the problem in the code seems correct. The LaTeX has some minor issues and the test can be made more robust. See my comments, mostly minor changes.

PR #548, opened 2026-06-26
Adding CTP1 problem

Ukez1:
Added the first test problem of the CTP-family.


gialmisi:
Good! The implemented CTP1 was correctly done. I removed two lines of dead code and implemented the rest.

PR #550, opened 2026-06-29
Cumulus

juropo:


desdeo/tools/group_scalarization.py
```diff
@@ -346,8 +346,8 @@ def add_group_asf_diff(
         name="alpha",
         symbol="_alpha",
         variable_type=VariableTypeEnum.real,
-        lowerbound=-float("Inf"),
-        upperbound=float("Inf"),
+        lowerbound=None,
+        upperbound=None,
```
gialmisi, line 350: I guess these changes are to make solvers happier? Or what was the original reason? Not opposing it, and it is more in line with how 'unbounded' variables are defined elsewhere as well.

juropo: That is exactly the reason.

I believe somewhere (in the documentation or something) it said that this was the canonical way of doing it.

desdeo/tools/partial_scalarization.py
```diff
@@ -91,75 +240,279 @@ def add_asf_partial_diff(
...
+
+    if not objectives:
+        return "0"
+    terms = []
+    for obj in objectives:
+        w = weights[obj.symbol]
+        if corrected_rp_aug is not None and obj.symbol in corrected_rp_aug:
+            terms.append(f"({obj.symbol} - {corrected_rp_aug[obj.symbol]}) / ({w})")
+        else:
+            terms.append(f"({obj.symbol}) / ({w})")
```
gialmisi: I think this will result in a sign bug for objectives to be maximized, unless I am missing something. The augmentation should be in minimization space, so this will work correctly for objectives to be minimized, but now for maximized ones. You could use `{obj.symbol}_min` here to fix this. Or define `obj_term = f"(-1 * {obj.symbol})" if obj.maximize else obj.symbol` at the top of the function and use it in the rest of the body. Whatever works.

desdeo/tools/partial_scalarization.py
```diff
@@ -91,75 +240,279 @@ def add_asf_partial_diff(
         if missing_nadir:
             msg = f"weights not provided and nadir is missing for objectives: {missing_nadir}."
             raise ScalarizationError(msg)
-        # Compute corrected nadir - corrected ideal in the minimization space.
         effective_weights = {obj.symbol: abs(obj.nadir - obj.ideal) for obj in active_objectives}
```
gialmisi: Regarding the comment on lines 304-309, here for example, the weights will always be positive and the missing "maximization-flip" is not cancelled, so this will result in a wrong-signed augmentation term. (affect both `add_asf_partial` and `add_asf_partial_nondiff` at least)

desdeo/tools/partial_scalarization.py
```diff
@@ -66,8 +67,156 @@ def add_asf_partial_diff(
...
+        symbol="_alpha",
+        variable_type=VariableTypeEnum.real,
+        lowerbound=None,
+        upperbound=None,
+        initial_value=1.0,
+    )
+
+    aug_expr = _build_aug_expr(
+        weights_aug, active_objectives, effective_weights, list(problem.objectives), corrected_rp_aug
+    )
```
gialmisi, line 90: Check comments for `_build_aug_expr`

desdeo/tools/partial_scalarization.py
```diff
@@ -66,8 +67,156 @@ def add_asf_partial_diff(
...
+
+    # Build the max term over active objectives only.
+    max_operands = []
+    for obj in active_objectives:
+        obj_term = f"(-1 * {obj.symbol})" if obj.maximize else obj.symbol
+        max_operands.append(f"({obj_term} - {corrected_rp[obj.symbol]}) / ({effective_weights[obj.symbol]})")
+    max_term = f"{Op.MAX}({', '.join(max_operands)})"
+
+    aug_expr = _build_aug_expr(
```
gialmisi, line 189: Check comment for `_build_aug_expr`

desdeo/tools/partial_scalarization.py
```diff
@@ -91,75 +240,279 @@ def add_asf_partial_diff(
...
-                f"(({obj.symbol}_min - {corrected_rp_aug[obj.symbol]}) / {aug_weights[obj.symbol]})"
-                for obj in active_objectives
-            ]
-        )
-
-    target_expr = f"_alpha + {rho}*({aug_expr})"
+    default_aug_weights = {obj.symbol: obj.nadir - (obj.ideal - delta) for obj in active_objectives}
+    aug_expr = _build_aug_expr(
+        weights_aug, active_objectives, default_aug_weights, list(problem.objectives), corrected_rp_aug
+    )
```
gialmisi: This happens to work currently because `default_aug_weights` is signed so the denominator's sign cancel the missing maximization-flip.

desdeo/problem/gurobipy_evaluator.py
```diff
@@ -368,7 +368,7 @@ def add_objective(self, obj: Objective):
             warnings.warn(
                 "One or more of the problem objectives seems to be a constant.", GurobipyEvaluatorWarning, stacklevel=2
             )
-        if isinstance(gp.GenExpr, int):
+        if isinstance(gp_expr, gp.GenExpr):
```
gialmisi, line 371: Lol, this was basically just checking the class instead of the object previously. Might explain some bugs I have had in the past. (nothing to correct here, this is a good catch)

desdeo/problem/utils.py
```diff
@@ -243,3 +252,126 @@ def tensor_constant_from_dataframe(
...
+        new_variables = [*new_variables, gte_var]
+        new_violation_symbols.append(gte_violation_symbol)
+
+    # Build the modified constraint func with violation variables absorbed
+    if constraint.cons_type == ConstraintTypeEnum.LTE:
+        modified_func = ["Subtract", constraint.func, lte_violation_symbol]
+    else:  # EQ
+        modified_func = ["Add", ["Subtract", constraint.func, lte_violation_symbol], gte_violation_symbol]
+
+    modified_constraint = constraint.model_copy(update={"func": modified_func})
```
gialmisi, line 339: The softened constraint keeps the original `constraint.symbol` here. Combined with the append-only logic at line 367, this is causes issue with constraints sharing the same symbol. If replacement is intended, the original same-symbol constraint needs to be removed here. Note that using `model_copy` does not re-validate the model, which is why no errors are raised here. I added the test case `test_add_soft_constraint_on_already_present_constraint_keeps_symbols_unique` in `test_problem_utils.py` which currently fails and showcases the duplicate symbol bug this can lead to.

desdeo/mcdm/cumulus.py
```diff
@@ -0,0 +1,606 @@
...
+        A 4-tuple ``(slack_vars, modified_soft_constraints, violation_sum_func, optimal_violation)``
+        where *slack_vars* are the new :class:`Variable` objects, *modified_soft_constraints* are
+        the slackened :class:`Constraint` objects, *violation_sum_func* is the MathJSON expression
+        for the sum of slacks, and *optimal_violation* is the minimized total violation value.
+    """
+    slack_vars: list[Variable] = []
+    modified_soft: list[Constraint] = []
+    for sc in soft_constraints:
+        slack_sym = f"_slack_{sc.symbol}"
+        slack_vars.append(Variable(name=f"Slack for {sc.name}", symbol=slack_sym, lowerbound=0.0, initial_value=0.0))
```
gialmisi: Missing `Variable` is missing the `variable_type` arg. This line would never execute.

desdeo/mcdm/cumulus.py
```diff
@@ -0,0 +1,606 @@
...
+        violation_sum_func: list = ["Multiply", 1, slack_vars[0].symbol]
+    else:
+        violation_sum_func = ["Add", *[sv.symbol for sv in slack_vars]]
+
+    viol_problem = problem
+    if hard_constraints:
+        viol_problem = viol_problem.add_constraints(hard_constraints)
+    viol_problem = viol_problem.add_variables(slack_vars)
+    viol_problem = viol_problem.add_constraints(modified_soft)
+    viol_problem = viol_problem.add_scalarization_func(
```
gialmisi: After fixing the above, the code would crash here next as there is not such method as `Problem.add_scalarization_func`. But `Problem.add_scalarization` does exist.

desdeo/mcdm/cumulus.py
```diff
@@ -0,0 +1,606 @@
...
+                cons_type=ConstraintTypeEnum.LTE,
+                func=["Subtract", violation_sum_func, optimal_violation],
+            )
+        ]
+    )
+
+    return {sf: solve_one(relaxed_problem, sf) for sf in scalarizations}
+
+
+def _minimize_soft_constraint_violations(
```
gialmisi, line 323: This function should have at least one test since, as of now, it will not execute without errors.

desdeo/mcdm/cumulus.py
```diff
@@ -0,0 +1,606 @@
...
+        solver_options (SolverOptions | None, optional): optional options passed
+            to the `solver`. Ignored if `solver` is `None`.
+            Defaults to None.
+
+    Returns:
+        SolverResults: the starting point solution.
+    """
+    ideal = problem.get_ideal_point()
+    nadir = problem.get_nadir_point()
+    if None in ideal or None in nadir:
```
gialmisi: Notice that this checks whether the key `None` is in is in ideal or nadir. But `Problem.get_ideal/nadir_point()` returns a dict which is always `{symbol: value or None}`. (the docstring was incorrect, but the typehints were correct in these methods.)

So this will always evaluate to `False`.

desdeo/mcdm/cumulus.py
```diff
@@ -0,0 +1,606 @@
...
+    classifications = {}
+
+    for obj in active_objectives:
+        rp_val = reference_point[obj.symbol]
+
+        # For minimized objectives the ideal/nadir are in the same space as rp_val.
+        # Guard against an aspiration better than ideal (would cause solver infeasibility).
+        if not obj.maximize and rp_val < obj.ideal:
+            classifications |= {obj.symbol: ("<", None)}
+            continue
```
gialmisi, line 174: What about maximized objectives here?

desdeo/mcdm/cumulus.py
```diff
@@ -0,0 +1,606 @@
...
+    if None in ideal or None in nadir:
+        msg = "The given problem must have both an ideal and nadir point defined."
+        raise CumulusError(msg)
+
+    if reference_point is None:
+        reference_point = {}
+    for obj in problem.objectives:
+        if obj.symbol not in reference_point:
+            reference_point[obj.symbol] = ideal[obj.symbol]
+
```
gialmisi: If the caller ever intends to re-use the same reference point supplies, this will mutate the the passed `reference_point`. Using a copy here instead would avoid unintended side-effects.

desdeo/mcdm/__init__.py
```diff
@@ -20,6 +22,7 @@
     "voting_procedure",
 ]
 
+from .cumulus import CumulusError, CumulusScalarization
```
gialmisi, line 29: Is there a particular reason only these two are exported from the module?

desdeo/mcdm/cumulus.py
```diff
@@ -0,0 +1,606 @@
...
+def _apply_scalarization(
+    problem: Problem,
+    sf: "CumulusScalarization",
+    scalarization_options: dict | None,
+    current_objectives: dict[str, float],
+    reference_point: dict[str, float],
+) -> tuple[Problem, str]:
+    """Apply one CUMULUS scalarization to *problem* and return the augmented problem and target symbol."""
+    match sf:
```
gialmisi, line 258: Consider adding a fallback case, e.g., `case _: raise CumulusError(...)`.

desdeo/mcdm/cumulus.py
```diff
@@ -0,0 +1,606 @@
...
+        ScalarizationFunction(
+            name="Total soft constraint violation",
+            symbol="_total_violation",
+            func=violation_sum_func,
+        )
+    )
+
+    viol_inst = init_solver(viol_problem, solver_options) if solver_options else init_solver(viol_problem)
+    viol_result = viol_inst.solve("_total_violation")
+    optimal_violation: float = viol_result.scalarization_values["_total_violation"]
```
gialmisi: FYI, this might be `None`.

desdeo/mcdm/cumulus.py
```diff
@@ -0,0 +1,606 @@
...
+    obj_by_sym = {obj.symbol: obj for obj in problem.objectives}
+    new_vars = dict(result.optimal_variables)
+    new_objs = dict(result.optimal_objectives)
+    new_extra = dict(result.extra_func_values or {})
+    new_scal = dict(result.scalarization_values or {})
+
+    for var in problem.variables:
+        t_sym = var.symbol
+        if not t_sym.startswith("_t_") or not var.name.startswith("Worst-case robust epigraph variable"):
+            continue
```
gialmisi: Check the current `add_single_objective_worst_case_regret`, it names its variable as "Worst-case regret epigraph...", which is a (I assume unintended) mismatch.

tests/test_cumulus.py
```diff
@@ -0,0 +1,336 @@
...
+
+    assert res_partial is not None and res_partial.success
+    assert res_pinned is not None and res_pinned.success
+
+    f2_partial = res_partial.optimal_objectives["f_2"]
+    f2_pinned = res_pinned.optimal_objectives["f_2"]
+
+    # The pinned solution must satisfy f_2 >= current (stay as good or better in min-space).
+    # In practice the two solutions should differ.
+    assert abs(f2_partial - f2_pinned) > 1e-6 or True  # structural: no crash; solver may agree
```
gialmisi: This is always true?

desdeo/api/routers/utils.py
```diff
@@ -372,3 +755,48 @@ def _validate(self, context: SessionContext) -> None:
...
+    """Collects all solutions for the user and problem."""
+    statement = (
+        select(StateDB)
+        .where(StateDB.problem_id == problem_id, StateDB.session_id == user.active_session_id)
+        .order_by(StateDB.id.desc())
+    )
+    states = session.exec(statement).all()
+    all_solutions = []
+    for state in states:
+        for i in range(state.state.num_solutions):
```
gialmisi, line 827: This would have crashed for `CumulusModificationState` previously, but now that it inherits (in my to be pushed PR) `ResultInterface`, this should not be a problem.

webui/src/routes/interactive_methods/CUMULUS/+page.svelte
```diff
@@ -0,0 +1,1188 @@
...
+					: 'Select final solution and finish the CUMULUS method with it'}
+			>
+				<Button
+					onclick={selectedIndexes.length === 1 ? confirm_finish : undefined}
+					disabled={selectedIndexes.length !== 1 || current_state.response_type === 'cumulus.finalize'}
+					variant="destructive"
+					class="ml-10"
+				>
+					Finish
+				</Button>
```
gialmisi, line 827: The starting point solution can now be finalized, which will break the backend's finalize endpoint.

webui/src/routes/interactive_methods/CUMULUS/handlers.ts
```diff
@@ -0,0 +1,482 @@
...
+	CumulusDeleteSaveRequest,
+	CumulusFinalizeRequest,
+	IntermediateSolutionRequest,
+	SolutionInfo,
+	CumulusObjectiveConstraintRequest,
+	SolutionDescriptionRequest,
+	ConstraintTypeEnum
+} from '$lib/gen/endpoints/DESDEOFastAPI';
+import type { ProblemInfo, Solution } from '$lib/types';
+import type { Response, ReferencePoint } from './types';
```
gialmisi: You should be able to use the Orval models directly here for some of these. If you look into `types.ts`, you will see that some of the imports are just referencing the generated Orval models. The reason these are still there is basically for "backwards" compatibility with the implementation of some UIs.

webui/src/routes/interactive_methods/CUMULUS/+page.svelte
```diff
@@ -0,0 +1,1188 @@
...
+			pending_scenario_model_id = result.scenario_model_id!;
+			pending_objective_symbols = result.objective_symbols!;
+			selected_measures = ['expected_value', 'worst_case_robust'];
+			cvar_alpha = 0.95;
+			combined_problem_name = problem ? `${problem.name} (combined)` : '';
+			show_scenario_setup_dialog = true;
+			return;
+		}
+
+		apply_initialization_result(result);
```
gialmisi: This should probably be awaited.

gialmisi (changes requested):
Pushed minor changed to your branch in [2169241](https://github.com/industrial-optimization-group/DESDEO/pull/550/commits/21692411df37b94014ab9706c63b395ac29ac79e). Be sure so pull it. 

Things that are still likely best addressed:

- Soft constraint relaxation is never reached. `_minimize_soft_constraint_violations` in mcdm/cumulus.py crashes twice: Variable is built without variable_type, and it calls a non existent `Problem.add_scalarization_func`. Reachable from `/solve` with objective constraints when every subproblem is infeasible.
- Augmentation sign is flipped for maximized objectives. `_build_aug_expr` in `tools/partial_scalarization.py` uses the raw objective instead of its minimization form, so the aug term is wrong signed in `add_asf_partial_diff/nondiff` and in the `reference_point_aug` path. The scenario path hides it via signed weights, as I understand it.
- `generate_starting_point` guard never triggers. `None` in ideal checks dict keys, not values. It should be `None` in `ideal.values()`.
- `finalize` assumes a list. `solver_results[idx]` breaks for init and save states, reachable from the UI by finalizing the starting point (which is possible right now in the UI).
- `add_soft_constraint` does not dedupe an already present constraint. Regression test added in my commit, and it currently fails. This should be addressed.
- `infer_classifications` lacks the maximize guard for a reference point better than ideal.

desdeo/tools/partial_scalarization.py
juropo: Yeah, I believe you are correct. I think the augmentation reference points might not used anywhere right now, so I didn't notice this.

desdeo/problem/utils.py
juropo: It was intended that the constraint should not already exist in the problem. But it's a good idea to add a check for that.

desdeo/mcdm/__init__.py
juropo: Yes, there is. I forgot to update the init

juropo: Actually, should any of the method-specific functions be imported here? There's a bunch of functions with conflicting names in the different nautilus variants

desdeo/mcdm/cumulus.py
juropo: Just like with NIMBUS, the symbol does not actually change whether the object is maximized or not.

desdeo/mcdm/cumulus.py
juropo: Good point

desdeo/mcdm/cumulus.py
juropo: It should be less bad now. Although, I assume I will find some bugs once I eventually add the UI.

desdeo/mcdm/cumulus.py
juropo: That is a different thing that probably needs its own fix function

desdeo/mcdm/cumulus.py
juropo: This and the other bug were copied from NIMBUS. Better fix those as well, I suppose

desdeo/tools/partial_scalarization.py
juropo: I think I fucked this up when consolidating repeating code into functions

tests/test_cumulus.py
juropo: Yeah, that's a bad test. Must have been for some debugging purpose and then left like that.

webui/src/routes/interactive_methods/CUMULUS/+page.svelte
juropo: yes

webui/src/routes/interactive_methods/CUMULUS/handlers.ts
juropo: yes

webui/src/routes/interactive_methods/CUMULUS/+page.svelte
juropo: Same problem exists in NIMBUS but I fixed it here

juropo:
Now, I think everything should be in order.

desdeo/mcdm/cumulus.py
gialmisi: Fair enough.

gialmisi:
@juropo Checked the changes and fine by me. Noticed an f-string missing the identifier, but otherwise I think my comments that were worth addressing were addressed. Feel free to merge.

PR #553, opened 2026-07-03
Add ADM Afsar/Chen implementations and test suite

Ardarito:
- Add test_adm.py: unit tests covering the ADM (Artificial Decision Maker) workflow with the RPM solver.

- Update ADMChen.py:

* Fix payoff_table_method return handling: the function returns dicts 
    {objective_name: value}, not np.ndarray, causing TypeError in arithmetic 
    operations (zstar - eps). Values are now extracted and converted to arrays.

 * Fix chained comparison validation in generate_initial_preference(): 
    NumPy arrays don't support Python-style chained comparisons 
    (a <= b <= c), which raised "ValueError: the truth value of an array 
    is ambiguous". Replaced with two separate np.all() calls.

- Update ADMfsar.py:

(Note: ADMChen.py is still WIP — several aspects of the original Chen et al. (2023) article are ambiguous and need further clarification)

* Fix broadcasting error in generate_reference_point_decision(), 
    generate_ranges_decision(), and generate_preferred_solutions_decision(): 
    max_assigned_vector can be an array of shape (N,) when there's a tie 
    between reference vectors, causing incompatible broadcasting against 
    assigned_vectors of shape (20,). Now uses max_assigned_vector[0] to 
    select the first winning index consistently across all three methods.

- Update how_to_utilize_mcdm_methods.ipynb: add usage example demonstrating the ADM + RPM interactive loop.


Ardarito:
## Additional changes (commits ad06bba and 993f77e)

On top of the original ADMAfsar implementation, the following was added:

- Replaced Euclidean distance with an ASF (achievement scalarizing function)
  performance measure when selecting solutions within each subpopulation
  assigned to a reference vector, making it comparable to the classic
  `rpm_solve_solutions`.

- Exposed `reference_vectors_` and `assigned_vectors_` as public attributes,
  visible on every call to `get_next_preference`.

- Used `generate_initial_preference()` directly, removing the redundant
  second call to `payoff_table_method`.

- Bug fix: applied `min()`/`max()` per axis before calling `rng.uniform()`,
  since some problems internally store the ideal/nadir points using a
  maximization convention where `ideal[i] > nadir[i]`, which previously
  caused a `ValueError: high < low`.

- New module `rpm_evolutionary.py`: an evolutionary version of RPM
  (RVEA/NSGA-III) for the k+1 sub-problems, exposing `assigned_vector` per
  sub-problem while keeping the same output interface as
  `rpm_solve_solutions`.

PR #572, opened 2026-08-18
Aggregating reference points fairly in NAUTILI

jpajasmaa:
This is not yet published work on how to aggregate reference points of several DMs in NAUTILI context fairly. What is fair is defined according to Rawlsian Maxmin fairness and local preference models, additive and symmetric cones, are used to evaluate how well each DM is satisfied with any given candidate group reference point.


gialmisi (commented):
Seems fine to me. Just a comment about the nautili.md example. Consider transforming it into a notebook (check exsiting guides, they are actually notebooks that are automagically converted to markdown). This way, the code in the example is executed when the docs are building, and if it fails, we will know. Its basically an extra layer of testing, and also forces us to publish executable code in the docs.

Otherwise feel free to merge.

PR #575, opened 2026-08-24
Core-logic, tournament selection seed fix

gialmisi:
Seed tournament selection so NSGA-II reproduces.The options constructor passed no seed, so mate selection drew from an unseeded generator. Passing it alone would have switched the selection from deterministic to stochastic, since `seed` controlled both, so the mode is now an optional `deterministic` flag that defaults to the old inference.


gialmisi:
Closing as already addressed...

PR #577, opened 2026-08-25
docs: fix OpenShift solver verification order

### LINKED ISSUE #576: Docs: Fix chronological order of solver verification in OpenShift deployment guide
**Background**
While following the OpenShift deployment guide (deploying_on_openshift.md) to set up hosting on Rahti, I noticed a minor issue regarding the execution order of the steps. For (special) users following the commands strictly top-to-bottom in-order, a step fails because a resource hasn't been created yet.

**Description**
In Step 4: Trigger first builds, immediately after the oc start-build instructions, the guide tells the user to verify the solvers inside the running pod:

`oc exec deployment/desdeo-api -- which bonmin ipopt cbc`

However, if the guide is followed chronologically (specially), the API deployment has not actually been created on the cluster yet. That happens in Step 5: Deploy API and web UI.

Running the exec command in Step 4 results in the following error:
`
Error from server (NotFound): deployments.apps "desdeo-api" not found `

**Suggested fix is to move the solver verification down to Step 5** or not let special people to read the docs specially (probably more difficult).

mori-ikuri:
## Summary

- Move solver verification until after the API and web UI deployments are applied and their rollouts complete.
- Prevent the documented top-to-bottom flow from failing with `NotFound` before `deployment/desdeo-api` exists.

## Validation

- Changed-file pre-commit hooks passed.
- MkDocs build passed.

Fixes #576

### DIFF
--- a/docs/howtoguides/deploying_on_openshift.md
+++ b/docs/howtoguides/deploying_on_openshift.md
@@ -316,14 +316,6 @@ The first build takes longer than subsequent ones because there is no layer
 cache. Expect roughly a few minutes for the builder, and anther few minutes for
 both the API and the webui.
 
-Once the API pod is running, verify the solvers are present:
-
-```bash
-oc exec deployment/desdeo-api -- which bonmin ipopt cbc
-```
-
-All three should return paths under `/opt/solver_binaries/`.
-
 !!! warning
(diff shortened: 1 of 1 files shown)


gialmisi:
Thank you, well spotted. Merged.

PR #585, opened 2026-09-03
Fix RVEA and CompositeTerminator under evaluation budgets

gialmisi:
- Fixed RVEASelector crashing on a generation with no feasible member, since the ideal was updated from an empty selection.
- Made CompositeTerminator forward messages to its children and expose their topics, so that a limit inside a composite counts. Children the publisher already delivers to are skipped.
- Made RVEASelector adapt its reference vectors periodically under the function evaluation based strategy, instead of once.
- Added tests for all three.


gialmisi:
@light-weaver I ran into some issues trying to solve constrained problems with RVEA. These are minor changes, but I am not 100% sure if it changes how the method works in a fundamental way.

desdeo/emo/operators/selection.py
```diff
@@ -716,12 +718,20 @@ def do(
...
+            # ideal yet, take it over every member, as the unconstrained branch does;
+            # that is optimistic, which an ideal point may be, and it is replaced as
+            # soon as a feasible member appears.
+            if feasible.any():
+                feasible_ideal = np.min(targets[feasible], axis=0)
+                self.ideal = (
+                    feasible_ideal if self.ideal is None else np.min(np.vstack((self.ideal, feasible_ideal)), axis=0)
+                )
+            elif self.ideal is None:
+                self.ideal = np.min(targets, axis=0)
```
light-weaver, line 734: Unfortunately, the ideal of infeasible solutions might be much better than feasible ones. And that wrong ideal will propagate to all future generations according to line 731.

You can set some kind of boolean in line 734 that keeps track of whether the ideal is built from infeasible solutions. Then, the moment even a single feasible solution exists, replace the wrong ideal with the feasible_ideal instead of taking min(self.ideal, feasible_ideal) and switch the boolean. In all future generations, you can take the min as in 731 without any issue.

desdeo/emo/operators/termination.py
```diff
@@ -203,6 +203,36 @@ def __init__(self, terminators: list[BaseTerminator], publisher: Publisher, mode
             self.max_evaluations = max(max_evaluations)
         self.mode = mode
 
+    @property
+    def interested_topics(self):
```
light-weaver: This shouldn't be needed. The individual terminators should be subscribed to the publisher anyway (they would not be able to initialize without a publisher). No need to pass the information through the composite

desdeo/emo/operators/termination.py
```diff
@@ -203,6 +203,36 @@ def __init__(self, terminators: list[BaseTerminator], publisher: Publisher, mode
...
+        their behalf and forwards.
+        """
+        topics = []
+        for terminator in self.terminators:
+            for topic in terminator.interested_topics:
+                if topic not in topics:
+                    topics.append(topic)
+        return topics
+
+    def update(self, message: Message) -> None:
```
light-weaver: Again, this shouldn't be needed if the individual terminators are all subscribed to the publisher.

light-weaver: The only thing we potentially, eventually, need to fix is the state being sent to the publisher every time notify() is called. Currently I am suppressing all the internal states, and only sending the composite's state. This is fine for now because only the BaseTerminator handles the state for every terminator anyway. If that ever changes in the future, we would need to make sure that CompositeTerminator takes the union of the internal states

tests/test_ea_config.py
```diff
@@ -167,4 +176,84 @@ def test_ibea_river():
...
+    solver, extras = algorithms.emo_constructor(problem=problem, emo_options=options)
+    _ = solver()
+
+    assert extras.archive is not None and extras.archive.selections is not None
+    assert len(extras.archive.selections) > 0
+
+
+@pytest.mark.ea
+def test_an_evaluation_limit_inside_a_composite_terminator_binds():
+    """The template subscribes only top-level components, so a composite must forward.
```
light-weaver, line 213: not true

desdeo/tools/patterns.py
```diff
@@ -36,14 +37,34 @@ class that must be inherited by the classes that want to receive (or send) messa
...
-from abc import ABC, abstractmethod
+from abc import ABC, ABCMeta, abstractmethod
 from collections.abc import Sequence
 from typing import Literal
 
 from desdeo.tools.message import AllowedMessagesAtVerbosity, Message, MessageTopics
 
 
-class Subscriber(ABC):
+class _SubscriberMeta(ABCMeta):
```
light-weaver, line 47: This is the primary change. All other files have been changed to account for this.

Basically, no need to call auto_subscribe() or register_topics() manually anymore

tests/test_ea_config.py
light-weaver: Actually was true, now is fixed.

PR #597, opened 2026-09-19
District heating: JINA interactive methods (single-scenario, multi-scenario, two-stage robustness)

Gsaboori:
This work adds three interactive method user interfaces for a district heating, together with the API behind them.

**Where the data comes from:**

No problem data is included here. The problems are built from a decision maker's own model modules,
located through `DH_DATA_DIR`, and are then persisted in the database.

Multi-scenario and two-stage robustness **solve live** from the persisted problem and scenario model.
Single-scenario instead looks designs up in a **precomputed candidate pool** (`summary.csv` and
`pairwise_transfer_matrix_long.csv`, optionally `design_id_lookup.csv`) in the same folder; it never
solves. Registration attaches both metadata rows, so all three methods work after one run.

Solving uses Gurobi, so a deployment needs a licence valid on the server.

The model and data are not in this PR. I can send them privately to anyone who wants to run the methods, just let me know.


juropo:
I guess this goes without saying, but no one should merge this.

gialmisi:
@Gsaboori Thanks! I am always happy to see contributions. I will be checking this, and we can then discuss next steps. Currently, @juropo is correct, when it comes to the present state of the PR, which needs to be trimmed considerably and made more generic. I will try and support you with this.

Gsaboori:
> @Gsaboori Thanks! I am always happy to see contributions. I will be checking this, and we can then discuss next steps. Currently, @juropo is correct, when it comes to the present state of the PR, which needs to be trimmed considerably and made more generic. I will try and support you with this.

Thank you Giovanni.

